PR #277, opened 2026-05-07
Conformance sweep A4: validator extensions (6 ADR-driven checks)

davidlabianca:
## Conformance sweep A4: validator extensions (6 ADR-driven checks)

Closes #268

## Summary

- **6 ADR-driven validator/linter extensions** spanning `scripts/hooks/riskmap_validator/validator.py`, `scripts/hooks/riskmap_validator/graphing/graph_utils.py`, `scripts/hooks/precommit/_prose_tokens.py`, `scripts/hooks/precommit/validate_yaml_prose_subset.py`, `scripts/hooks/validate_framework_references.py`, and `scripts/hooks/validate_riskmap.py`.
- **Mode breakdown** (corrected against live-corpus inspection 2026-05-04):
  - **1 block-mode-immediately** (current corpus already passes): 2.3.7 lifecycle-order uniqueness
  - **4 warn-only-first with `--block` toggle** (current corpus has content debt; flips at sweep-closing commit per plan § 2.7.1): 2.3.5 folded-bullet drift live-corpus pinning, 2.3.6 deprecated-persona reference toggle, 2.3.8 controls↔components mirror, 2.3.9 category/subcategory nesting
  - **1 runtime warning** (operator-visible during generator runs; not a CI gate): 2.3.10 `MermaidConfigLoader` missing-category warning
- **Hybrid scope on 2.3.5** — A3 (PR #262) already shipped the folded-bullet detection heuristic in `_prose_tokens.py`. A4's contribution narrowed to (a) live-corpus regression pinning so issue #225 cases stay caught after future tokenizer edits; (b) augmented `INVALID_LIST` diagnostic message cross-referencing ADR-020 D4 + issue #225 so authors aim the right fix; (c) source/parsed-form documentation in the heuristic docstring. 
- **Live corpus debt surfaced by warn-only-first checks** (verified 2026-05-04; tracked separately for Phase B disposition):
  - 88 deprecated-persona refs (49 controls + 39 risks)
  - 2 dangling control→component refs (`componentInputHandling`/`componentOutputHandling` — likely typos for `componentApplication{Input,Output}Handling`)
  - 7 components missing `subcategory` field
- **Unified `--block` toggle** on `riskmap_validator/validator.py` covers tasks 2.3.8 + 2.3.9 via a shared `_warn_block_triggered` boolean — all warnings print before exit. One flip-point at § 2.7.1, not two.
[...]

PR #282, opened 2026-05-11
feat(precommit): codify trigger-vs-read-set invariant (#279)

### LINKED ISSUE #279: Codify hook-trigger ⊇ check-input ⊇ read-set invariant (ADR-005 addendum) + fix two trigger-vs-validator-scope gaps
## Summary

PR #211/#222 (bash → pre-commit framework migration, ADR-005) introduced a class of regression that PR #277's reviewer surfaced: a hook's `files:` regex can be narrower than what the validator actually reads, silently skipping validation on commits that don't touch the regex's match set. One hook is currently broken under this pattern; ADR-005 did not codify the invariant that would prevent recurrence.

Item 2 of the PR #277 review (lifecycle-stage trigger gap) landed in #277 itself via the new dedicated `validate-lifecycle-stage` hook. This issue tracks:

1. **Item 1**: widening `validate-component-edges` so controls-only or risks-only commits exercise the validator's full read set.
2. **Pre-existing trigger** for `validate-persona-site-build` correctly includes `controls.yaml` (added 2026-04-21 with the hook's introduction in commit `93cc22b`); add a regression test asserting controls-only commits trigger the hook so the trigger can't drift narrower in the future.
3. **ADR-005 addendum**: codify the invariant that the same architectural review surfaced.
4. **Structural test**: enforce the invariant for every local validator hook.

## Parent / context

- **Parent**: ADR-005 — Adopt the `pre-commit` framework for hook orchestration
- **Surfaced by**: PR #277 (`feature/A4-validator-extensions`); reviewer shrey-bagga's CHANGES_REQUESTED, 2026-05-08T02:07:59Z, items 1 + 2.
- **Related (already addressed)**: PR #277 commit `e8f1edb` adds `validate-lifecycle-stage` (item 2 of review).
- **Branch target**: `main`.

## Root cause

Pre-#211, the legacy bash harness invoked `validate_riskmap.py` unconditionally on every commit (see `scripts/hooks/pre-commit:411` in the deleted bash hook, removed by commit `463ecfd`). The script's own `get_staged_yaml_files()` decided what to scan from `[components.yaml, controls.yaml, risks.yaml]`.

Post-#211, `.pre-commit-config.yaml`'s `validate-component-edges` hook filters on `^risk-map/yaml/components\.yaml$` only. A commit that edits only `controls.yaml` or only `risks.yaml` does not trigger the hook, so the validator never runs — and none of its checks (component-edge consistency, A4's controls↔components mirror, A4's category/subcategory nesting) fire.

The same risk shape exists in `validate-persona-site-build` (the builder reads `controls.yaml`), but the trigger has correctly included `controls.yaml` since the hook's introduction in commit `93cc22b` (2026-04-21). The hook is therefore not currently broken; adding a regression test locks the correct behavior in so future edits to the trigger can't silently narrow it.

These are the *only* two affected hooks. A sweep of `.pre-commit-config.yaml` local-validators block confirmed all other hooks have either matching trigger/read sets or use `pass_filenames: true` (which couples scope to the filter by construction).

## The invariant

For every local validator hook in `.pre-commit-config.yaml`, with `pass_filenames: false`:

> The hook's `files:` regex MUST be a superset of every YAML path the
> validator reads, AND the validator's discovery surface
> (`get_staged_yaml_files()` `target_files` plus any direct file opens)
> MUST include those same paths. Equivalently: trigger-set ⊇ check-input-set
> AND read-set ⊇ check-input-set, for every check declared inside the
> validator.

The three sets:

- **Trigger set** — what the framework matches against staged files to decide whether to invoke the hook.
- **Read set** — what the validator opens on disk during a run.
- **Check-input set** — for each declared check, the YAMLs whose change would alter the check's verdict.

Item 1's gap is *trigger ⊉ read set* (validator already reads `controls.yaml`; only the trigger is narrow). Item 2's gap (already fixed in PR #277) was *trigger ⊉ check-input AND read ⊉ check-input* (`lifecycle-stage.yaml` wasn't in the validator's universe at all pre-A4). One invariant covers both. `validate-persona-site-build` is the third in-scope hook the structural test covers; its trigger has been correct since the hook's introduction, so it required only a regression-guard test (Task 2) rather than a `.pre-commit-config.yaml` change.

### 1. Widen `validate-component-edges` trigger

In `.pre-commit-config.yaml`, change:

```yaml
files: ^risk-map/yaml/components\.yaml$
```

to:

```yaml
files: ^risk-map/yaml/(components|controls|risks)\.yaml$
```

Validator's `target_files` already includes all three; only the trigger needs widening. Add a regression test asserting that a controls-only or risks-only staged change triggers the hook.

### 2. Regression guard for `validate-persona-site-build` trigger

Add a regression test asserting that a controls-only staged change triggers the persona-site build hook.

### 3. ADR-005 addendum

Author an addendum (or follow-up note) under `docs/adr/005-pre-commit-framework.md` codifying the invariant above. Place under § Decision (as a new bullet on hook authoring) or § Consequences → Negative (as a learned constraint), per architect's recommendation. Cross-reference this issue and PR #277.

The addendum should:

- State the invariant in the form given above.
- Note that `pass_filenames: true` couples trigger to scope by construction, so the invariant applies primarily to hooks with `pass_filenames: false`.
- Note the structural test (task 4) as the enforcement mechanism.
- Link the two regression incidents (PR #277 review items 1 + 2) as the motivating evidence.

### 4. Structural test enforcing the invariant

Add a new test class to `scripts/hooks/tests/test_precommit_hook_install.py` that, for every local validator hook with `pass_filenames: false`:

- Loads the hook's declared trigger regex.
- Looks up the validator's declared read set (either by parsing the validator module's `target_files` constant, or by maintaining a parallel metadata table keyed by hook id — the architect recommended a metadata table to avoid coupling the test to validator internals).
- Asserts trigger-set ⊇ read-set.

Document the metadata-table approach in the test's docstring so the next hook author knows where to register their hook's read set. Provide a clear failure message that names the hook id, the missing path, and a pointer to ADR-005 § <addendum-ref>.

## Out of scope

- Refactoring or extracting the inline lifecycle check at `validate_riskmap.py:184-220` (the default-mode belt-and-suspenders pattern PR #277 left in place). That's a separate code-quality concern.
- Auditing upstream hooks (`check-jsonschema`, `prettier`, `ruff`) for the same pattern — they don't apply (those hooks operate on the file list passed by the framework; they don't read additional files).
- Generator hooks (`regenerate-graphs`, `regenerate-tables`, `regenerate-svgs`) — they read multiple YAMLs and stage outputs; the trigger-superset rule may apply but the architecture is side-effectful so the invariant statement may need refinement before they're folded in. Defer to a follow-up if needed.

## Acceptance criteria

- [ ] `validate-component-edges` `files:` regex covers `(components|controls|risks)\.yaml`.
- [ ] `validate-persona-site-build` regression test asserts controls-only staged change triggers the hook.
- [ ] ADR-005 addendum landed; status flipped to Accepted by maintainer.
- [ ] Structural test passes; deliberately breaking either hook's trigger regex causes the test to fail with the named-hook + missing-path diagnostic.
- [ ] Pre-commit + pytest green across the full hook suite.
- [ ] Regression tests cover controls-only commit triggering both `validate-component-edges` (fixed by Task 1) and `validate-persona-site-build` (guarded by Task 2)

davidlabianca:
## feat(precommit): codify trigger-vs-read-set invariant (#279)

Closes #279                                                
                            
Implements the four tasks tracked at issue #279, surfaced by shrey-bagga's CHANGES_REQUESTED on PR #277:  
        
1. Widen `validate-component-edges` trigger to `(components|controls|risks)\.yaml`  
2. Regression guard for `validate-persona-site-build` trigger (the trigger has correctly included `controls.yaml` since the hook's introduction in commit `93cc22b`; this PR locks the behaviour in)
3. ADR-005 addendum codifying `trigger-set ⊇ check-input ⊇ read-set` invariant with three carve-outs (generators / master-schema fan-out / runtime-discovered read sets)  
4. Structural test in `test_precommit_hook_install.py` (`TestTriggerReadSetInvariant`) iterates every local `pass_filenames: false` hook against a `_LOCAL_VALIDATOR_READ_SETS` metadata table; drift-prevention test  
asserts every such hook is registered  
       
Diff: 3 files, +524/-2. Single commit `06e15a0`. 
                                                  
## Test plan       
                                  
- [x] `pytest scripts/hooks/tests/test_precommit_hook_install.py` — 32 passed  
- [x] Full hook suite — `pytest scripts/hooks/tests` — 2329 passed / 6 skipped (pre-rebase; current corpus unchanged)  
- [x] `ruff check` + `ruff format --check` clean         
- [x] `pre-commit run --all-files` clean  
- [x] Deliberately breaking either widened trigger causes the structural test to fail with the named-hook + missing-path diagnostic (architect-reviewed)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -208,7 +208,7 @@ repos:
         name: 'risk-map: component edge consistency'
         language: system
         entry: python3 scripts/hooks/validate_riskmap.py
-        files: ^risk-map/yaml/components\.yaml$
+        files: ^risk-map/yaml/(components|controls|risks)\.yaml$
         pass_filenames: false
 
       - id: validate-control-risk-references
--- a/docs/adr/005-pre-commit-framework.md
+++ b/docs/adr/005-pre-commit-framework.md
@@ -109,7 +109,7 @@ For every local validator hook in [`.pre-commit-config.yaml`](../../.pre-commit-
 Three sets, defined per hook:
 
 - **Trigger set** — the file paths the framework matches against staged files (the hook's `files:` regex) to decide whether to invoke the hook.
-- **Read set** — the file paths the validator opens on disk during a run (its `target_files` discovery surface plus any direct opens).
+- **Read/discovery set** — the fixed file paths the validator treats as in scope during a run, including `target_files` staged-file discovery surfaces and direct file opens. This can be broader than the files literally parsed on the default hook path when a validator preserves a legacy discovery surface.
 - **Check-input set** — for each declared check inside the validator, the file paths whose contents could change the check's verdict.
 
 Hooks declared with `pass_filenames: true` couple their trigger to their scope by construction — the framework hands the validator the matched file list and the validator works on exactly that input — so the invariant primarily targets `pass_filenames: false` hooks, where trigger and read-set are independent declarations that can drift apart silently.
@@ -126,13 +126,13 @@ Refining the invariant for these classes is deferred to a separate decision.
 
 ### Enforcement
 
-The structural test at [`scripts/hooks/tests/test_precommit_hook_install.py`](../../scripts/hooks/tests/test_precommit_hook_install.py) enforces the invariant for every in-scope local validator hook. Each new hook registers its read set in a metadata table keyed by hook id (architect-recommended over parsing validator internals), so the next hook author has one declared place to record what their validator actually reads. A test failure names the hook id, the missing path, and points back to this addendum.
+The structural test at [`scripts/hooks/tests/test_precommit_hook_install.py`](../../scripts/hooks/tests/test_precommit_hook_install.py) enforces the invariant for every in-scope local validator hook. Each new hook registers its fixed trigger coverage set in a metadata table keyed by hook id (architect-recommended over parsing validator internals), so the next hook author has one declared place to record the paths that must cause the hook to run. A test failure names the hook id, the missing path, and points back to this addendum.
 
 ### Motivating regressions
 
 Two trigger-vs-read-set gaps inherited from the bash-to-framework migration ([#211](https://github.com/cosai-oasis/secure-ai-tooling/pull/211) / [#222](https://github.com/cosai-oasis/secure-ai-tooling/pull/222)) surfaced in review of [PR #277](https://github.com/cosai-oasis/secure-ai-tooling/pull/277):
 
-- **`validate-component-edges`** — trigger filtered on `components.yaml` only, but the validator's read set already covered `controls.yaml` and `risks.yaml`. Controls-only or risks-only commits skipped the validator entirely.
-- **`validate-lifecycle-stage`** — `lifecycle-stage.yaml` was outside both the trigger and the read set pre-A4. Resolved within PR #277 itself by introducing a dedicated hook with a narrow trigger and a `--mode lifecycle` short-circuit on `validate_riskmap.py`.
+- **`validate-component-edges`** — trigger filtered on `components.yaml` only, but the validator's staged-file discovery surface already covered `controls.yaml` and `risks.yaml`; the default check path also parses `controls.yaml` for the controls↔components mirror check. Controls-only or risks-only commits skipped the validator entirely.
+- **`validate-lifecycle-stage`** — `lifecycle-stage.yaml` was outside both the trigger and the read/discovery set pre-A4. Resolved within PR #277 itself by introducing a dedicated hook with a narrow trigger and a `--mode lifecycle` short-circuit on `validate_riskmap.py`. The remaining default-mode lifecycle belt-and-suspenders check inside `validate_riskmap.py` is deliberately covered by that dedicated hook's contract, not by `validate-component-edges`.
 
 Both are documented in [#279](https://github.com/cosai-oasis/secure-ai-tooling/issues/279). The invariant exists to make the same class of regression a structural-test failure rather than a reviewer catch.
--- a/docs/adr/005-pre-commit-framework.md
+++ b/docs/adr/005-pre-commit-framework.md
@@ -104,7 +104,7 @@ Authored 2026-05-08; accepted alongside the implementation of [#279](https://git
 
 For every local validator hook in [`.pre-commit-config.yaml`](../../.pre-commit-config.yaml) with `pass_filenames: false`:
 
-> **trigger-set ⊇ check-input-set, AND read-set ⊇ check-input-set.**
+> **trigger-set ⊇ check-input-set, AND read/discovery-set ⊇ check-input-set.**
 
 Three sets, defined per hook:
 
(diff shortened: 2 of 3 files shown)


shrey-bagga:
I pushed a follow-up fix in 878bcb3 to address the trigger/read-set wording issue.

The important distinction is that `risks.yaml` is part of `validate-component-edges`' fixed staged-file discovery surface, but it is not literally parsed by the normal default hook path. The structural metadata now calls this a fixed trigger coverage set instead of a literal read set, keeps `risks.yaml` in scope per #279's acceptance criteria, and documents that `lifecycle-stage.yaml` is owned by the dedicated `validate-lifecycle-stage` hook rather than this hook's trigger contract.

Verification run on the PR branch:

- `/Users/sbagga/Desktop/CoSAI/secure-ai-tooling/.venv/bin/python -m pytest scripts/hooks/tests/test_precommit_hook_install.py` -- 32 passed
- `ruff check scripts/hooks/tests/test_precommit_hook_install.py` -- passed
- `ruff format --check scripts/hooks/tests/test_precommit_hook_install.py` -- passed

shrey-bagga (approved):
LGTM

shrey-bagga:
Clarification on the PR #282 review finding and commit `878bcb3`:

During review, we found that the new structural test and ADR were using the term “read set” in a confusing way.

For `validate-component-edges`, `risks.yaml` is supposed to trigger the hook, but the normal pre-commit validation path does not literally read `risks.yaml`. So the table is really describing which files must trigger the hook, not only which files the hook opens during normal validation.

Commit `878bcb3` updates the wording to make that clearer:
- it describes the table as fixed trigger coverage instead of a literal read set;
- it keeps `risks.yaml` in scope for PR #282’s intended behavior;
- it clarifies that `lifecycle-stage.yaml` is handled by the dedicated `validate-lifecycle-stage` hook.

davidlabianca:
> I pushed a follow-up fix in [878bcb3](https://github.com/cosai-oasis/secure-ai-tooling/commit/878bcb3227f764e55110a27c146bbeaf0baa5f94) to address the trigger/read-set wording issue.
> 
> The important distinction is that `risks.yaml` is part of `validate-component-edges`' fixed staged-file discovery surface, but it is not literally parsed by the normal default hook path. The structural metadata now calls this a fixed trigger coverage set instead of a literal read set, keeps `risks.yaml` in scope per #279's acceptance criteria, and documents that `lifecycle-stage.yaml` is owned by the dedicated `validate-lifecycle-stage` hook rather than this hook's trigger contract.
> 
> Verification run on the PR branch:
> 
> * `/Users/sbagga/Desktop/CoSAI/secure-ai-tooling/.venv/bin/python -m pytest scripts/hooks/tests/test_precommit_hook_install.py` -- 32 passed
> * `ruff check scripts/hooks/tests/test_precommit_hook_install.py` -- passed
> * `ruff format --check scripts/hooks/tests/test_precommit_hook_install.py` -- passed

Thanks - good catch and the terminology switch helps the ADR.  I'm pushing a small polish commit to keep consistency with your reframing as read discovery set (ADR only)

PR #20, opened 2025-08-29
Initial contribution of the SAIF risk map data to CoSAI's secure-ai-tooling repo

davidlabianca:
Initial contribution of [Google's SAIF data](https://github.com/google/saif-data) to CoSAI as a sub-project in `cosai-oasis/secure-ai-tooling`. 

Modifications from source fork:
- Updates to cover CoSAI naming and contribution guidelines
  - Includes a schema file name update and all references
- Updates to the `secure-ai-tooling/README.md` to cover the sub-project

PR #283, opened 2026-05-11
This commit does fix CodeQL #20 + #21 (redos in diagnostic-format regex)

davidlabianca:
## This commit does fix CodeQL 20 + 21 (redos in diagnostic-format regex)

Adds the dot to the inner negated character class in _DIAG_PATTERN across both prose-linter test files.  

The fragment `[^:\[]+(?:\.[^:\[]+)*` admitted exponential backtracking on inputs with many dotted segments when the trailing [<digit>] match failed (~53s on N=30 segments). The fix `[^:\`[.]+(?:\.[^:\[.]+)*` makes the dot exclusively the segment separator, eliminating the matching ambiguity.  

Adds a regression test in each file asserting bounded match time on the redos seed.

__Note: the production linters only emit the diagnostic string via f-string — they never parse it with a regex. _DIAG_PATTERN exists only in the test suites as a format-contract assertion, so the issues are correctly scoped to the test files and nothing else needs fixing.__

### DIFF
--- a/scripts/hooks/tests/test_validate_prose_references.py
+++ b/scripts/hooks/tests/test_validate_prose_references.py
@@ -150,7 +150,7 @@ _REPO_ROOT = Path(__file__).resolve().parent.parent.parent.parent
 # The references linter embeds the token value directly in reason strings
 # (e.g. "... at 'riskBaz'"); the regex matches any non-empty reason.
 # ---------------------------------------------------------------------------
-_DIAG_PATTERN = re.compile(r"^validate-prose-references: [^:]+:[^:]+:[^:\[]+(?:\.[^:\[]+)*\[\d+\]: .+$")
+_DIAG_PATTERN = re.compile(r"^validate-prose-references: [^:]+:[^:]+:[^:\[.]+(?:\.[^:\[.]+)*\[\d+\]: .+$")
 
 # ---------------------------------------------------------------------------
 # Helpers for building synthetic YAML and schema fixtures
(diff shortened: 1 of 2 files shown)


shrey-bagga:
## Code Review: PR #283

**Status:** ITERATE

**Summary:** The regex change addresses the CodeQL ReDoS finding cleanly by making `.` an exclusive field-segment separator, and the affected test files pass locally. The only change I recommend before merge is strengthening the new regression tests so they fail cleanly if the vulnerable regex shape comes back.

**Required changes**

| Severity | Location | Problem | Recommended fix |
| --- | --- | --- | --- |
| minor | `scripts/hooks/tests/test_validate_prose_references.py` / `scripts/hooks/tests/test_validate_yaml_prose_subset.py` ReDoS tests | The tests assert elapsed time only after `_DIAG_PATTERN.match(...)` returns. If this regresses to catastrophic backtracking, CI can stall before reaching the assertion. | Add `@pytest.mark.timeout(2)` or equivalent to both ReDoS regression tests. `pytest-timeout` is already part of the project requirements, so this aligns with the repo's test tooling and contributor expectation that regressions fail deterministically. |

**Suggestions:** None beyond the timeout guard.

**Questions:** None.

**Next steps:** Route back to the PR author/testing for the timeout annotations, then this should be ready for maintainer review.

**Local verification:**

- `pytest scripts/hooks/tests/test_validate_prose_references.py scripts/hooks/tests/test_validate_yaml_prose_subset.py` -> 156 passed
- `ruff check scripts/hooks/tests/test_validate_prose_references.py scripts/hooks/tests/test_validate_yaml_prose_subset.py` -> passed

scripts/hooks/tests/test_validate_prose_references.py
```diff
@@ -1658,3 +1658,39 @@ def test_camelcase_diagnostic_reason_contains_snippet(self):
...
+
+# ===========================================================================
+# TestDiagnosticPatternRedosResistance
+# ===========================================================================
+
+
+class TestDiagnosticPatternRedosResistance:
+    r"""CodeQL #21 regression guard for the diagnostic-format regex."""
+
+    def test_diagnostic_pattern_resists_redos(self):
```
shrey-bagga, line 1672: Add a pytest-timeout guard so a future ReDoS regression fails instead of hanging CI.

```suggestion
    @pytest.mark.timeout(2)
    def test_diagnostic_pattern_resists_redos(self):
```

scripts/hooks/tests/test_validate_yaml_prose_subset.py
```diff
@@ -1160,6 +1160,33 @@ def test_dotted_field_name_in_diagnostic_format(self, tmp_path):
         line = self._diag_to_line(diag)
         assert _DIAG_PATTERN.match(line), f"Diagnostic line does not match pattern: {line!r}"
 
+    def test_diagnostic_pattern_resists_redos(self):
```
shrey-bagga, line 1164: Add a pytest-timeout guard so a future ReDoS regression fails instead of hanging CI.

```suggestion
    @pytest.mark.timeout(2)
    def test_diagnostic_pattern_resists_redos(self):
```

shrey-bagga (commented):
Suggested fix for the timeout guard noted in the code review.

shrey-bagga (approved):
Small nit pick. LGTM though

scripts/hooks/tests/test_validate_yaml_prose_subset.py
davidlabianca: implemented in 47251f2

PR #309, opened 2026-05-14
Add local generated table drift checking

### LINKED ISSUE #301: tooling: `validate-all.sh` doesn't flag table-vs-YAML drift; CI catches it only after push
### Summary

`scripts/tools/validate-all.sh` runs every risk-map validator with `--force` but **explicitly skips regeneration of tables, graphs, and SVGs** (per its own docstring at line 7: *"without regenerating graphs, tables, or SVGs"*). This is by design for fast iteration loops.

The consequence: a maintainer can edit `risk-map/yaml/*.yaml` during a rebase, forget to run the table regenerator, and have `validate-all.sh` report green — because no validator checks whether `risk-map/tables/**.md` matches what the generator would produce from the current YAML.

CI **does** catch this after rebase push — `.github/workflows/validate_tables.yml` ("Validate Markdown Tables") runs regeneration and compares against committed tables, failing on `mismatched_files`. 

Net effect: maintainer pushes, CI goes red, round-trip cost.

## Reproduction (exemplar)

Surfaced during the 2026-05-14 Phase B rebase cycle (B2 / 284.1 rebase paths) where conflict resolution touched YAML but pre-commit's `regenerate-tables` hook didn't fire as part of every step:

1. Resolve a YAML conflict during rebase (e.g., `risks.yaml`)
2. `git add risk-map/yaml/risks.yaml` without re-running pre-commit
3. Run `./scripts/tools/validate-all.sh` — passes (all validators green)
4. `git rebase --continue` succeeds
5. Push → CI `Validate Markdown Tables` workflow fails because `risks-full.md` no longer matches what `yaml_to_markdown.py` produces from the current `risks.yaml`

### Hard purity contract (non-negotiable)

The new check MUST run without **any** side effects on the working tree or repo state. Specifically:

- **No mutation of tracked files.** `risk-map/tables/**`, `risk-map/diagrams/**`, `risk-map/svg/**`, and all YAML, schema, and source files MUST be byte-identical before and after a run (success, failure, or interrupted).
- **No writes outside a designated temp directory.** All regenerated artifacts go to `$(mktemp -d)` (or equivalent). The temp path is the only writable surface.
- **Temp directory cleaned up on exit.** Cleanup MUST run on success, failure, and signal interrupt (SIGINT / SIGTERM). Implement via `trap '<cleanup>' EXIT INT TERM` so the cleanup is unconditional. Acceptable to leave debris if the script is SIGKILL'd; that's the OS limit, not a tooling concern.
- **No git operations.** The check MUST NOT invoke `git add`, `git commit`, `git checkout`, or any state-changing git command. `git diff` for reading is fine; nothing that mutates the index or working tree.
- **No environment-variable persistence.** Any env vars set for the regenerator subprocess MUST be local to that subprocess (no exporting to caller's shell).
- **Exit code is the only persistent signal.** `0` on parity; non-zero on drift or error.

### Verification leverage

`scripts/hooks/yaml_to_markdown.py` already accepts `--output-dir`. The implementation lever is:

```bash
yaml_to_markdown.py --all --all-formats --output-dir "$TMPDIR/tables" --quiet
```

This writes the canonical table set to `$TMPDIR/tables/` without touching `risk-map/tables/`. The check then compares (via `diff -r`, byte-shasum, or per-file `cmp`) the two directories and reports drift.

`scripts/hooks/precommit/regenerate_tables.py` is **NOT** a suitable entry point — it writes directly to `risk-map/tables/` and runs `git add` as a side effect (lines 38-45, 82, 100, 126, 144). The verification path bypasses it entirely.

### Implementation shape (recommended)

Add an opt-in flag to `scripts/tools/validate-all.sh`:

```bash
./scripts/tools/validate-all.sh --check-generation
```

Pseudocode for the new mode:

```bash
TMPDIR="$(mktemp -d)"
trap 'rm -rf "$TMPDIR"' EXIT INT TERM
mkdir -p "$TMPDIR/tables"

# Regenerate to temp (no side effects on tracked tree)
python3 scripts/hooks/yaml_to_markdown.py --all --all-formats \
    --output-dir "$TMPDIR/tables" --quiet

# Diff against committed
if ! diff -r -q risk-map/tables "$TMPDIR/tables" > "$TMPDIR/drift.txt"; then
    echo "Table drift detected:" >&2
    cat "$TMPDIR/drift.txt" >&2
    # Optional: show first N lines of unified diff for each drifted file
    exit 1
fi
```

Default behavior of `validate-all.sh` (called without `--check-generation`) is unchanged. Flag is opt-in.

### Naming

`--check-generation` is the recommended flag name. Alternatives considered:
- `--strict` — too vague; doesn't say what it's strict about
- `--with-regen` — implies it regenerates committed tables; opposite of the purity contract
- `--verify-tables` — narrow; doesn't extend to graphs/SVGs if scope expands
- `--check-generation` — accurate, future-proof if graphs/SVGs join later

## Acceptance criteria

- [ ] `validate-all.sh --check-generation` regenerates the table set to a temp location and reports any drift against `risk-map/tables/`
- [ ] Exit code `1` on any drift detected; exit code `0` only when committed artifacts exactly match generator output
- [ ] **Purity contract verified by automated test**:
  - [ ] Tracked files unchanged after run (success, failure, and SIGINT cases)
  - [ ] Temp directory removed after each exit path
  - [ ] No `git` index mutation (verify `git status` is identical before/after)
- [ ] Default `validate-all.sh` behavior unchanged (no perf regression for fast-iteration callers without the flag)
- [ ] `--help` documents the new flag and its purity contract
- [ ] Drift report names each drifted file (not just exit-code yes/no)
- [ ] Docs updated in `risk-map/docs/contributing/` or `risk-map/docs/validation.md` to point at strict mode for pre-push verification
- [ ] CI invariant unchanged — the existing `validate_tables.yml` workflow remains the source of truth for the remote-side check

## Tests (mandatory for the purity contract)

The purity contract is the load-bearing part of this fix. Tests MUST cover:

1. **No tracked-file mutation on success path** — set up a clean tree, run `validate-all.sh --check-generation`, assert no `risk-map/**` or `scripts/**` file changed (compare git-tree hash before and after)
2. **No tracked-file mutation on drift path** — introduce synthetic YAML drift, run the flag, assert exit code 1 AND no tracked file changed (only the temp dir was written)
3. **Temp dir cleanup on success** — capture the temp path during the run (via debug hook or trace), assert it doesn't exist after the script exits
4. **Temp dir cleanup on failure** — force regenerator failure (e.g., malformed YAML), assert temp dir is still cleaned up
5. **Temp dir cleanup on SIGINT** — start the script in a subprocess, send SIGINT, assert temp dir is cleaned up
6. **No git index changes** — capture `git status --porcelain` before and after; must be byte-identical

## Out of scope

- Re-architecting the generators themselves — this is about adding verification, not changing generation
- Replacing pre-commit's `regenerate-tables` hook — pre-commit is the auto-stage path; `validate-all.sh` is the verification path
- Adding parity check to CI — CI already has `validate_tables.yml`; the gap is only on the local side
- Graphs (`risk-map/diagrams/**`) and SVGs (`risk-map/svg/**`) — these have the same gap pattern but a separate generator (`regenerate_graphs.py` / `regenerate_svgs.py`); if extending to those is desired, do it in a follow-on commit/PR after this one establishes the pattern for tables

## Related

- `scripts/tools/validate-all.sh` — the tool with the gap
- `scripts/hooks/yaml_to_markdown.py` — the generator (already supports `--output-dir`)
- `scripts/hooks/precommit/regenerate_tables.py` — the pre-commit wrapper (NOT suitable as a verification entry point — has side effects)
- `.github/workflows/validate_tables.yml` — the CI workflow that catches drift remotely (the contract this fix aligns with)
- Pattern surfaced during Phase B conformance sweep rebase cycles (B2 / 284.1, 2026-05-14)

shrey-bagga:
## Summary
- add opt-in validate-all.sh --check-generation for side-effect-free table drift checks
- compare temp-generated markdown tables against risk-map/tables without touching tracked files or the git index
- document the strict local pre-push verification path

Fixes #301

## Validation
- bash -n scripts/tools/validate-all.sh
- ruff check scripts/hooks/tests/test_validate_all_tool.py
- ruff format --check scripts/hooks/tests/test_validate_all_tool.py
- pytest scripts/hooks/tests/test_validate_all_tool.py
- ./scripts/tools/validate-all.sh --check-generation

### DIFF
--- a/risk-map/docs/validation.md
+++ b/risk-map/docs/validation.md
@@ -107,7 +107,8 @@ python3 scripts/hooks/yaml_to_markdown.py components --format summary
 
 This strict mode generates tables in a temporary directory, compares them with
 `risk-map/tables/`, and exits non-zero when drift is found. It does not update
-tracked files or change the git index.
+tracked files or change the git index. Any warnings the generator writes to
+stderr still surface in the output; only generator stdout is suppressed.
 
 **Available formats:**
 
--- a/risk-map/docs/validation.md
+++ b/risk-map/docs/validation.md
@@ -98,6 +98,17 @@ python3 scripts/hooks/yaml_to_markdown.py controls --format xref-risks
 python3 scripts/hooks/yaml_to_markdown.py components --format summary
 ```
 
+**Pre-push drift check:**
+
+```bash
+# Verify committed tables match generated output without modifying the tree
+./scripts/tools/validate-all.sh --check-generation
+```
+
+This strict mode generates tables in a temporary directory, compares them with
+`risk-map/tables/`, and exits non-zero when drift is found. It does not update
+tracked files or change the git index.
+
 **Available formats:**
 
 - `full` - Complete tables with all columns (default)
--- a/risk-map/docs/validation.md
+++ b/risk-map/docs/validation.md
@@ -107,8 +107,8 @@ python3 scripts/hooks/yaml_to_markdown.py components --format summary
 
 This strict mode generates tables in a temporary directory, compares them with
 `risk-map/tables/`, and exits non-zero when drift is found. It does not update
-tracked files or change the git index. Any warnings the generator writes to
-stderr still surface in the output; only generator stdout is suppressed.
+tracked files or change the git index. Generator errors print to stdout
+regardless of `--quiet`; only the generator's progress lines are suppressed.
 
 **Available formats:**
 
--- a/scripts/docs/manual-validation.md
+++ b/scripts/docs/manual-validation.md
@@ -39,6 +39,11 @@ the prior `pre-commit.sh --force` workflow.
 # Validate everything, no commit, no regeneration:
 ./scripts/tools/validate-all.sh
 
+# Validate everything and verify generated markdown tables are current.
+# This writes generated tables only to a temporary directory and does not
+# change tracked files or the git index:
+./scripts/tools/validate-all.sh --check-generation
+
 # Help:
 ./scripts/tools/validate-all.sh --help
 ```
--- a/scripts/hooks/tests/test_validate_all_tool.py
+++ b/scripts/hooks/tests/test_validate_all_tool.py
@@ -230,6 +230,26 @@ def test_check_generation_cleans_tempdir_on_sigint(tmp_path: Path):
     assert not (tmp_path / "git-invocations.log").exists()
 
 
+def test_check_generation_fails_cleanly_when_mktemp_fails(tmp_path: Path):
+    """A failing `mktemp -d` must surface as a clean validator failure, not
+    cascade into a write outside the temp tree (e.g. mkdir -p "/tables").
+    """
+    repo, env = _make_stubbed_repo(tmp_path)
+    before_status = _git_status(repo)
+
+    # Override mktemp on PATH so the script's `mktemp -d` returns non-zero
+    # with no stdout. This simulates a full TMPDIR or permission failure.
+    stub_bin = Path(env["PATH"].split(os.pathsep)[0])
+    _write_executable(stub_bin / "mktemp", "#!/bin/bash\nexit 1\n")
+
+    result = _run_validate_all(repo, env, "--check-generation")
+
+    assert result.returncode == 1
+    assert "Could not create temporary directory" in result.stdout
+    _assert_repo_unchanged(repo, before_status)
+    assert not (tmp_path / "git-invocations.log").exists()
+
+
 def test_help_documents_check_generation_purity_contract(tmp_path: Path):
     repo, env = _make_stubbed_repo(tmp_path)
 
--- a/scripts/hooks/tests/test_validate_all_tool.py
+++ b/scripts/hooks/tests/test_validate_all_tool.py
@@ -14,6 +14,8 @@ import subprocess
 import time
 from pathlib import Path
 
+import pytest
+
 REPO_ROOT = Path(__file__).parent.parent.parent.parent
 SCRIPT_SOURCE = REPO_ROOT / "scripts" / "tools" / "validate-all.sh"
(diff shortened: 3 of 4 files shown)


davidlabianca:
@shrey-bagga — The purity contract holds cleanly and the test suite maps to all 6 mandatory categories from #301. Pushed three small follow-up commits on top of `d894185` to land the notes:

- **`1c8db45`** — harden `validate-all.sh`: guard `mktemp -d` failure so an empty `GEN_TMPDIR` cannot cascade into a write outside the temp tree; document the `set -u`-without-`-e` invariant; note `check_generated_tables` owns the EXIT trap and that `exit 130` re-fires it idempotently; invert the `if/then/:/else` idiom to match the five sibling validators; soften the header to "by default, does not regenerate." Adds `test_check_generation_fails_cleanly_when_mktemp_fails`.
- **`92809b8`** — parametrize the signal-cleanup test over both SIGINT and SIGTERM (trap covers `INT TERM` but only INT was exercised); add `test_check_generation_flags_extra_file_in_tables_dir` as a regression guard for `diff -r -q`'s file-set behavior; harden the python3 stub to accept both `--output-dir <DIR>` and `--output-dir=<DIR>` forms.
- **`1cb523d`** — pre-push tightening: replace a misleading stderr claim in `validation.md` (the generator actually writes errors to stdout, not stderr; `--quiet` only gates progress); collapse the `mktemp` guard rationale to the single non-obvious line.

9/9 tests pass; `bash -n` + ruff clean; smoke-tested end-to-end. CI invariant unchanged (no `.github/workflows/` touched).

davidlabianca (approved):
lgtm - small nit commits with rationale in comment

PR #340, opened 2026-05-22
Add schema registry coverage to persona site data tests

### LINKED ISSUE #307: Latent-bug audit: `jsonschema.validate()` calls without `referencing.Registry` resolver
## Latent-bug audit: `jsonschema.validate()` calls without `referencing.Registry` resolver

**Surfaced by:** B2 PR #305 review (2026-05-14). Companion to commit `8670fdd` (test fix for `test_persona_site_data_schema_matches_generated_output`).

### Problem

`scripts/hooks/tests/test_build_persona_site_data.py::test_persona_site_data_schema_matches_generated_output` failed during B2 once `risks.yaml` gained populated `externalReferences` arrays. The root cause was a `$ref` to `external-references.schema.json` in `persona-site-data.schema.json` that the test's `jsonschema.validate()` call had no resolver for.

The bug was latent — present in the test long before B2, but invisible because no entity in the live corpus had populated `externalReferences`. Any data point that traversed the `$ref` path triggered it; absence of such data silently passed.

### Latent at this sibling site

`scripts/hooks/tests/test_build_persona_site_data.py::test_persona_site_data_schema_rejects_structural_violations` (lines 497-498) also calls `jsonschema.validate()` without a registry. Currently safe because the mutators it parametrizes against only inject structural violations (missing/wrong-type fields), not data that traverses the `externalReferences` `$ref`. A future mutator that populates `externalReferences` would silently regress to the same `Unresolvable` exception.

### Recommendation

Audit every `jsonschema.validate()` call in the test tree. Where the validated schema `$refs` another schema, pass `registry=_make_schema_registry(risk_map_schemas_dir)` (the production helper at `scripts/build_persona_site_data.py:47`). The test in `test_persona_site_data_schema_matches_generated_output` is the reference pattern after B2's `8670fdd`.

Suggested approach:
1. Grep `scripts/hooks/tests/` for every `jsonschema.validate(` call site
2. For each, identify whether the schema (and its referenced schemas) contain `$ref` 
3. Audit: does the test data traverse those `$ref` paths under any parametrization?
4. If yes (or could in future), add the registry

### Severity

HIGH — silent passing. Tests that "pass" because they don't exercise the bug pattern are not actually validating what they advertise. The pattern is a class of bug, not a one-off.

### Why no B2-scope action

The B2 PR's scope is content (anchor migrations) + the single test fix surfaced by content. Expanding to a tree-wide audit dilutes the PR's review surface and isn't required to land B2. Filing as follow-up.

## Related

- B2 PR #295 (this work)
- ADR-016 D3 — `externalReferences` schema
- Companion follow-up: promote `_make_schema_registry()` from private to shared helper

shrey-bagga:
## Summary

- Add the persona site data schema registry to the structural-violation `jsonschema.validate()` test.
- Add a cross-file schema-reference invalid `externalReferences` case so the regression is pinned.
- Audit every validator call in `scripts/hooks/tests/` against the class of bug #307 describes (validator runs against a schema with cross-file `$ref` without a `referencing.Registry`).

Closes #307.

## Root cause

Issue #307 identified a latent sibling test that validated `persona-site-data.schema.json` without a `referencing.Registry`. That was safe only while its mutators avoided fields backed by cross-file schema references — a future mutator touching `externalReferences` (or any other cross-file `$ref` field) would have raised `referencing.exceptions.Unresolvable`, which is *not* a subclass of `jsonschema.ValidationError` and would not be caught by the `pytest.raises(jsonschema.ValidationError)` assertion.

## What changed

`scripts/hooks/tests/test_build_persona_site_data.py` (+34 / −2):

1. `test_persona_site_data_schema_rejects_structural_violations` now builds `registry = _make_schema_registry(risk_map_schemas_dir)` and passes `registry=registry` into `jsonschema.validate()` — the same resolver path the live-output validation already uses.
2. A new parametrize case (`external-references-url-non-https`) appends a risk whose `externalReferences[0].url = "http://example.com/reference"` to traverse the cross-file `$ref` and exercise the `^https://` constraint that lives only in `external-references.schema.json`.

The mutator is load-bearing: without the registry the assertion would fail (Unresolvable, not ValidationError); with the registry it correctly raises ValidationError on the non-HTTPS URL.

## Class-of-bug audit (tree-wide)

Per #307's framing — *"the pattern is a class of bug, not a one-off"* — every `jsonschema`-family validator call in `scripts/hooks/tests/` was surveyed for the same latent shape (validator runs against a schema with cross-file `$ref` without a `referencing.Registry`). Two call-site forms were checked: `jsonschema.validate(...)` (top-level) and `Draft7Validator(schema).validate(...)`.

| Call site | Wrapped schema | `registry=` | Cross-file `$ref` traversed? |
|---|---|---|---|
| `test_build_persona_site_data.py:376` | `persona-site-data.schema.json` (whole) | yes (pre-existing, `8670fdd`) | yes — covered |
| `test_build_persona_site_data.py:551` | `persona-site-data.schema.json` (whole) | **yes (this PR)** | yes — covered |
| `test_lifecycle_stage_order_range.py:78` | `lifecycle-stage` `order` property | no | no (leaf int schema) |
| `test_riskmap_prose_strict.py:96,102` | `riskmap.schema.json#/definitions/utils/{text,prose-strict}` | no | no (self-contained) |
| `test_framework_mapping_patterns.py:324,348,428` | `framework-mapping-patterns` entries | no | no (string+pattern leaves) |
| `test_external_references_schema.py:147,290,303` | `external-references.schema.json` fragments | no | no (file is self-contained) |
| `test_persona_site_data_prose_items.py:128` | `persona-site-data.schema.json#/definitions/prose` | no | no today; fragile (see follow-up) |

**Net:** the two `jsonschema.validate(` sites identified in #307 were the only sites whose target schema actually traverses a cross-file `$ref` at validation time. Both are now registry-aware. One forward-fragility note (`test_persona_site_data_prose_items.py:128`) is filed as a separate LOW follow-up.

## Validation

- `python3 -m pytest scripts/hooks/tests/test_build_persona_site_data.py -q` — full file passes
- `python3 -m pytest scripts/hooks/tests/test_build_persona_site_data.py -q -k structural_violations -v` — 4 / 4 parametrized cases fail-as-expected
- `ruff check scripts/hooks/tests/test_build_persona_site_data.py` — clean
- `./scripts/tools/validate-all.sh --quiet` — clean
- `python3 -m pytest scripts/hooks/tests -q` — 2403 passed, 9 skipped (two local-environment-sensitive failures unrelated to this PR: Playwright Chromium dry-run skip assertion, and verify-deps.sh requiring Python >= 3.14 while this machine has 3.12.10)

### DIFF
--- a/scripts/hooks/tests/test_build_persona_site_data.py
+++ b/scripts/hooks/tests/test_build_persona_site_data.py
@@ -490,8 +490,36 @@ risks:
             ),
             "persona object missing required 'id' field",
         ),
+        (
+            lambda d: d["risks"].append(
+                {
+                    "id": "riskWithExternalReference",
+                    "title": "Risk With External Reference",
+                    "category": "risksTest",
+                    "shortDescription": [],
+                    "longDescription": [],
+                    "examples": [],
+                    "controlIds": [],
+                    "personaIds": [],
+                    "externalReferences": [
+                        {
+                            "type": "other",
+                            "id": "non-https-ref",
+                            "title": "Non-HTTPS reference",
+                            "url": "http://example.com/reference",
+                        }
+                    ],
+                }
+            ),
+            "risk object with externalReferences violating the shared referenced schema",
+        ),
     ],
-    ids=["missing-top-level-key", "wrong-type-top-level", "nested-object-missing-id"],
+    ids=[
+        "missing-top-level-key",
+        "wrong-type-top-level",
+        "nested-object-missing-id",
+        "referenced-external-references-schema",
+    ],
 )
 def test_persona_site_data_schema_rejects_structural_violations(
     risk_map_schemas_dir: Path,
@@ -512,11 +540,15 @@ def test_persona_site_data_schema_rejects_structural_violations(
     with schema_path.open("r", encoding="utf-8") as handle:
         schema = json.load(handle)
 
+    # Keep this rejection test on the same resolver path as live-output
+    # validation, so future mutators can safely traverse cross-file $refs.
+    registry = _make_schema_registry(risk_map_schemas_dir)
+
     bad = copy.deepcopy(_minimal_valid_site_data())
     mutator(bad)
 
     with pytest.raises(jsonschema.ValidationError):
-        jsonschema.validate(instance=bad, schema=schema)
+        jsonschema.validate(instance=bad, schema=schema, registry=registry)
 
 
 # ============================================================================


davidlabianca:
Thanks @shrey-bagga ...

1. Amended PR body to land the tree-wide audit on record (independent architect + code-reviewer pass). The diff itself is unchanged 

   - the two registry-aware `jsonschema.validate(` sites are still the only call sites whose schema traverses a cross-file `$ref` at validation time
   - the new `external-references-url-non-https` mutator is a real regression guard (verified: without the registry the assertion would raise `Unresolvable`, not `ValidationError`, and would fail rather than silently pass).

2. Two non-blocking nits for future follow-up remain unaddressed in the diff:

   - `test_build_persona_site_data.py:521` — id `referenced-external-references-schema` names the mechanism, not the defect; `external-references-url-non-https` would match the other three ids' pattern.
   - `test_build_persona_site_data.py:514` — description string has the same shape; `risk object with non-HTTPS URL in externalReferences` would be parallel to the other three descriptions.

3. One LOW-priority follow-up filed separately (#360) for a forward-fragility note in `test_persona_site_data_prose_items.py`: line 128 wraps a self-contained sub-schema today but the same file at line 405 uses a registry — mixed convention in one file is a foot-gun if `definitions/prose` ever gains a `$ref`. Not a bug today; just a consistency cleanup.

4. The `_make_schema_registry()` promotion the issue suggested as a companion follow-up stays parked — only one test caller exists (this PR doesn't add new callers), and promotion gains nothing until a third caller emerges.

davidlabianca (approved):
lgtm

PR #350, opened 2026-05-25
tooling: harden sanitizer structured-item rendering

### LINKED ISSUE #308: tooling: sanitizer hardening — fail-loud on unknown structured-item types + escapeLinkTitle rename
### Summary

Two related sanitizer-module observations surfaced by the code-reviewer during B2 PR #305 validation (2026-05-14). Neither is a regression today — both protect future-proofing posture and code clarity. Bundled here because both touch `site/assets/sanitizer.mjs` and both are small.

#### What

`site/assets/sanitizer.mjs:431-434` (the post-`71ae880` shape) falls back to `escapeHtml(String(input))` when `renderProse` receives an item it doesn't recognize:

```js
const escaped = escapeHtml(String(input));  // for {type: "alien", ...} → "[object Object]"
console.warn("renderProse: escaped unexpected markup", { input, escaped });
return escaped;  // emits literal "[object Object]" into the DOM
```

The code-reviewer smoke-tested this with `{type: "alien", foo: "bar"}` and confirmed: output is exactly the regression string the B2 effort eliminated. The `console.warn` fires, so it isn't completely silent — but CI doesn't assert on console output, and the rendered DOM carries the same `[object Object]` symptom the bleed-thru gate was designed to catch.

#### Why it matters

**If a future builder change adds a third structured-item type** without simultaneously extending `renderProse`, the exact regression class B2 just fixed silently returns — but this time on the new type, not on `ref`. The fix would still be a one-line dispatch addition; the problem is that nothing surfaces the gap until someone manually checks rendered output.

This is a fail-loud gap, not a current bug. Today there are exactly two structured-item types (`link` + `ref`); both are handled. The hardening is forward-looking insurance.

#### Fix options

**Option A1 — throw on unknown type:**

```js
throw new Error(`renderProse: unsupported structured-item type ${JSON.stringify(input.type)}`);
```

CI catches new-type additions automatically (existing tests fail; new tests have to be added before the dispatch can be extended). Forces fail-loud at the test boundary. Slight downside: production bugs that produce unknown types crash the renderer instead of degrading.

**Option A2 — emit HTML-inert marker:**

```js
return `<!-- renderProse: unsupported prose-item type: ${escapeHtml(String(input.type))} -->`;
```

Rendered output is searchable (`grep -r 'unsupported prose-item type'` in CI output or in the rendered site) but DOM-inert. Production degrades gracefully (nothing visible) while still leaving a forensic trail. The bleed-thru regression test from B2 already greps for `[object Object]` — extend it to also fail on the marker substring, gaining fail-loud behavior without crashing the renderer.

**Recommended: A2 with an extended regression test.** Combines fail-loud-in-CI with graceful-degrade-in-prod. Matches the bleed-thru gate's existing pattern (output-level regression scan).

#### Acceptance criteria (Finding A)

- [ ] Unknown structured-item types no longer produce literal `[object Object]` in rendered output
- [ ] CI regression test fails when an unknown type is introduced (assert on rendered substring or `console.error` count)
- [ ] Existing 30 sanitizer tests + 14 `app-render-rich-paragraphs` tests still green
- [ ] If Option A2 is chosen: bleed-thru regression in B2's existing test suite extended to grep for the marker substring as well

---

#### What

Per `71ae880`'s extension, the ref branch at `sanitizer.mjs:420` calls:

```js
return `<a href="#${escapeHtml(id)}">${escapeLinkTitle(String(input.title ?? ""))}</a>`;
```

Same `escapeLinkTitle` helper is used in the link branch. The function correctly handles both — ref titles are display text, not markdown, exactly like link titles. **Behaviorally correct.** But the function name now implies link-specific scope when it's actually structured-title-generic.

#### Why it matters

Code clarity for future readers who look at `escapeLinkTitle` and wonder whether ref titles need a different helper. The reviewer flagged it as "mildly misleading" — pure cosmetic, no behavioral impact.

#### Fix

Rename `escapeLinkTitle` → `escapeStructuredTitle` (or `escapeProseItemTitle`, or similar). Single-symbol rename; all call sites in `sanitizer.mjs` and `sanitizer.test.mjs` follow. Verify with `git grep escapeLinkTitle` after rename returns empty.

#### Acceptance criteria (Finding B)

- [ ] Helper renamed from `escapeLinkTitle` to `escapeStructuredTitle` (or maintainer-chosen alternative)
- [ ] All call sites updated (`sanitizer.mjs` + `sanitizer.test.mjs`)
- [ ] `git grep escapeLinkTitle` returns empty
- [ ] All site tests still green
- [ ] Helper's JSDoc (if any) reflects the post-rename surface

---

### Scope discipline

Both findings are scoped to `site/assets/sanitizer.mjs` and its tests. Nothing in `app.mjs`, the builder, or any YAML touched. Combined effort: ~1-2 hours including tests.

### Out of scope

- Restructuring the sanitizer module's architecture beyond these two changes
- Extending renderProse to handle a third structured-item type (no current builder emits one)
- Changes to `app.mjs` or `renderRichParagraphs` beyond inheriting from the helper rename if it affects them
- ADR amendments — the current ADR-015/ADR-016 text doesn't conflict with either fix; no addendum needed
- The original ref-item bleed (resolved in B2 PR #295 commit `71ae880`; see sibling tracker)

### Related

- B2 PR #295 — surfaced both findings during review (2026-05-14)
- Commit `71ae880` — renderProse ref extension (resolved the original ref bleed; surfaced Finding A's fail-loud gap)
- Sibling tracker: `gh-drafts/draft-b2-followup-ref-item-render-path.md` — closes-as-fixed-in-B2; provides historical context for the diagnostic path
- ADR-015 D1 — sanitizer module scope
- ADR-016 D5 (with 2026-05-14 addendum) — ref item rendering spec
- `site/assets/sanitizer.mjs` — module under modification
- `site/tests/sanitizer.test.mjs` + `site/tests/app-render-rich-paragraphs.test.mjs` — test surfaces

## Notes for contributor

Both findings are "good first issue" sized. Finding A is the more interesting one (forward-looking fail-loud posture); Finding B is mechanical. Either can be tackled independently — they share a file but not a code path. If you tackle Option A2 (emit marker), reuse the bleed-thru regression-test pattern already established in B2 PR #305 rather than inventing new test scaffolding.

The maintainer's preference is Option A2 for Finding A — degrades gracefully in production while remaining fail-loud in CI via the existing regression-scan pattern.

shrey-bagga:
## Summary

- Rename `escapeLinkTitle` to `escapeStructuredTitle` now that the helper is shared by link and ref structured items.
- Emit an HTML-inert unsupported-type marker for unknown structured prose items instead of rendering `[object Object]`.
- Extend sanitizer and rich-paragraph regression tests to catch unknown structured-item bleed-through.

## Why

Issue #308 identified a forward-looking sanitizer gap: if a future builder introduces a new structured prose item type before `renderProse` supports it, the fallback path could reproduce the `[object Object]` regression class. This keeps production output inert while making the marker easy for CI regression scans to catch.

Closes #308.

## Validation

- `node --test site/tests/*.test.mjs` -- 54/54 passing
- `node_modules/.bin/prettier --check site/assets/sanitizer.mjs site/tests/sanitizer.test.mjs site/tests/app-render-rich-paragraphs.test.mjs`
- `git diff --check`
- `git grep escapeLinkTitle` returned no matches

### DIFF
--- a/site/assets/sanitizer.mjs
+++ b/site/assets/sanitizer.mjs
@@ -21,6 +21,13 @@
 // this pattern exists. Adding a tag here requires a matching fixture update.
 const ALLOWED_TAGS = ["strong", "em", "a"];
 
+// Bounded-dispatch literal for structured-item types — same posture as
+// ALLOWED_TAGS. Adding a type here without extending renderProse's dispatch
+// arms OR without adding a per-type positive test fails the meta-tests in
+// sanitizer.test.mjs. This is the structural fail-loud guarantee for new
+// structured-item types alongside the per-call inert-marker fallback.
+const STRUCTURED_ITEM_TYPES = ["link", "ref"];
+
 // ---------------------------------------------------------------------------
 // HTML character escaping
 // ---------------------------------------------------------------------------
@@ -65,6 +72,29 @@ function escapeHtml(text) {
  */
 function escapeStructuredTitle(text) {
   return escapeHtml(text).replaceAll("=", "&#61;");
+}
+
+/**
+ * Reduce an unknown structured-item `type` value to a short label for the
+ * inert-marker diagnostic. Avoids two regression classes:
+ *   1. Non-string `type` (object/array) producing literal "[object Object]"
+ *      via String() coercion — the bleed-thru gate would trip on that.
+ *   2. Null bytes in a string `type` surviving into the rendered innerHTML
+ *      string; HTML5 parsers replace U+0000 with U+FFFD inside comments at
+ *      parse time, but the raw innerHTML carries the null and confuses
+ *      programmatic grep-based regression scans.
+ *
+ * @param {unknown} value - The `type` field of an unrecognised structured item
+ * @returns {string} Short label safe to feed through escapeHtml
+ */
+function describeUnsupportedType(value) {
+  if (typeof value === "string") {
+    return value.replace(/\x00/g, "\uFFFD");
+  }
+  if (value === null) return "<null>";
+  if (value === undefined) return "<undefined>";
+  if (Array.isArray(value)) return "<array>";
+  return `<non-string:${typeof value}>`;
 }
 
 // ---------------------------------------------------------------------------
@@ -428,9 +458,13 @@ export function renderProse(input) {
       return escaped;
     }
 
-    // Unknown structured type — emit an inert marker instead of leaking
-    // String(object) as "[object Object]" into rendered prose.
-    const escaped = `<!-- renderProse: unsupported prose-item type: ${escapeHtml(String(input.type))} -->`;
+    // Unknown structured type — emit an inert marker. describeUnsupportedType
+    // keeps the marker free of "[object Object]" (which would trip the bleed-
+    // thru gate when input.type is itself an object) and strips null bytes
+    // before escapeHtml so the raw innerHTML stays greppable. The marker is
+    // wrapped in an HTML comment; escapeHtml neutralises `>` so a hostile
+    // input.type cannot close the comment early.
+    const escaped = `<!-- renderProse: unsupported prose-item type: ${escapeHtml(describeUnsupportedType(input.type))} -->`;
     console.warn("renderProse: escaped unexpected markup", { input, escaped });
     return escaped;
   }
--- a/site/assets/sanitizer.mjs
+++ b/site/assets/sanitizer.mjs
@@ -1,8 +1,8 @@
 /**
  * sanitizer.mjs — render-time prose sanitizer for the CoSAI Risk Map site.
  *
- * Exports renderProse(input) which accepts either a plain string or a
- * structured link object ({type: "link", title, url}) and returns an
+ * Exports renderProse(input) which accepts a plain string, a structured link
+ * object ({type: "link", title, url}), or a structured ref object and returns an
  * HTML-safe string suitable for assignment to innerHTML.
  *
  * Design: ADR-015 D3 (hand-rolled, bounded emission), D3a (ALLOWED_TAGS
@@ -50,7 +50,7 @@ function escapeHtml(text) {
 }
 
 /**
- * Escape a string for use as anchor text content.
+ * Escape a string for use as structured-item anchor text content.
  * Extends escapeHtml by also encoding `=` as `&#61;` so that attribute-
  * injection patterns like `onclick=` cannot appear as a raw substring in the
  * rendered HTML output, even inside text content where they are already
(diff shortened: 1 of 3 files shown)


afiune (commented):
LGTM.

One small nitpick, since this is sanitizer-boundary code, it may be worth adding one adversarial
unknown-type fixture in addition to the benign `{ type: "alien" }` case. For example, an unknown
`type` containing `<script>`, quotes, or comment-like characters would lock in that the marker
stays inert and does not accidentally introduce raw HTML when the diagnostic includes `input.type`

davidlabianca:
> LGTM.
> 
> One small nitpick, since this is sanitizer-boundary code, it may be worth adding one adversarial unknown-type fixture in addition to the benign `{ type: "alien" }` case. For example, an unknown `type` containing `<script>`, quotes, or comment-like characters would lock in that the marker stays inert and does not accidentally introduce raw HTML when the diagnostic includes `input.type`

agreed - it is a good safety net

davidlabianca:
Pushed `72ac8e6` to address @afiune's adversarial-fixture nit and four related gaps surfaced by a follow-up parallel architect + code-reviewer pass. 60/60 site tests passing (was 54/54).

**Changes**

- **Adversarial-type sweep** (`sanitizer.test.mjs`) — 15-fixture parametrized test covering hostile `type` values: `<script>...</script>`, `-->`, `<!--`, `--!>`, `--`, quote injection, null byte, object/array/numeric/boolean/null `type`, missing key, and array-as-input. Each asserts exactly one `<!--`+`-->` pair, no raw `<script>`, no raw null byte, no `[object Object]`. Pins the load-bearing invariant that `escapeHtml` neutralises `>` so a hostile `input.type` cannot close the comment early.
- **`describeUnsupportedType` helper** (`sanitizer.mjs`) — coerces non-string `type` to short labels (`<null>`, `<undefined>`, `<array>`, `<non-string:typeof>`). Removes the second-order regression where `type: { nested: ... }` previously embedded literal `[object Object]` inside the marker (which would have tripped the bleed-thru gate). Also replaces `\x00` → `�` on string `type` so the raw innerHTML stays greppable for CI scans.
- **`STRUCTURED_ITEM_TYPES` registry** (`sanitizer.mjs`) — `const STRUCTURED_ITEM_TYPES = ["link", "ref"]` literal mirroring `ALLOWED_TAGS` (ADR-015 D3a). Three meta-tests in `sanitizer.test.mjs` enforce: (1) literal-const shape via source read, (2) each entry has a matching `input.type === "<type>"` dispatch arm in source, (3) each entry has at least one positive test case. Adding a third type without extending the dispatch *or* without a positive fixture now fails CI — extends the structural fail-loud posture from tag-emission to structured-item dispatch.
- **Bleed-thru gate mirrored at unit layer** — `assertNoStructuredItemBleedThrough` lifted into `sanitizer.test.mjs` (was previously only in `app-render-rich-paragraphs.test.mjs`) and applied to positive link + new positive ref tests. A regression that causes a known-type path to emit the unsupported-type marker now fails at both the unit and integration layers.

**Validation**

- `node --test site/tests/*.test.mjs` → 60/60
- `prettier --check` on all three touched files → clean
- `git grep escapeLinkTitle` → empty (B.3 still met)
- `pre-commit run --files` on touched files → all green

**Scope discipline**

Stays inside #308's stated bounds: `site/assets/sanitizer.mjs` + its tests only. No `app.mjs`/builder/YAML touched. No ADR amendment — the new posture sits inside ADR-015 D4's existing escape-with-warn frame; the registry mirrors D3a's literal-const discipline.

davidlabianca (approved):
LGTM

PR #363, opened 2026-05-27
Update components-full.md

smartpill212:
Swap the componentApplicationInputHandling description and componentApplicationOutputHandling descriptions to match the component names.


davidlabianca:
Thanks @smartpill212 - this is a real description bug but the fix is wrong. The correct update is to fix the source `yaml`. 

Closing and tracking in Issue #364

Issue #159, opened 2026-03-11
[infra]: Replace pip with uv for Python dependency management


totallyGreg:
## Enhancement Title
Replace pip with uv for Python dependency management

## Enhancement Category
Tooling / Scripts / Validation

## Overview
Replace `pip` with [`uv`](https://docs.astral.sh/uv/) as the Python package installer and environment manager across `install-deps.sh`, `verify-deps.sh`, and CI workflows. uv provides deterministic virtual environment creation, 10-100x faster installs, and eliminates the PATH/shim race conditions that cause pip packages to appear uninstalled after `install-deps.sh` completes.

## Rationale & Justification
**Current state:** `pip install -r requirements.txt` installs packages into a site-packages location that may not match what `python3 -m pip show` later checks, especially when mise shims aren't fully activated during script execution. This results in all 8 pip packages reporting as `[FAIL] not installed` during verification despite a successful install step.

**Proposed improvement:** Use `uv sync` with a project-local `.venv` for deterministic, isolated dependency management.

**Benefits:**
- Eliminates pip/mise shim PATH race conditions that cause false verification failures
- 10-100x faster package resolution and installation (significant for CI and devcontainer builds)
- Project-local `.venv` makes the environment reproducible and isolated
- mise has first-class `uv` support (`uv = "latest"` in `.mise.toml`)
- Aligns with modern Python standards (PEP 621 `[project.dependencies]` in `pyproject.toml`)
- Single tool replaces pip for both library installs and CLI tool management

## Scope
Medium (multiple files, moderate complexity)

### Phase 1: Add uv tooling
- Add `uv = "latest"` to `.mise.toml`
- Migrate `requirements.txt` dependencies into `[project.dependencies]` in `pyproject.toml`

### Phase 2: Update install and verify scripts
- Update `install-deps.sh` Step 4 to use `uv sync` instead of `pip install -r requirements.txt`
- Update `verify-deps.sh` pip package checks to use `uv pip show` or check `.venv` presence

### Phase 3: Update CI and container config
- Update CI workflows that reference pip
- Update devcontainer Dockerfile if pip is referenced
- Remove `requirements.txt` after migration

## Task List / Subtasks
- [ ] Add `uv = "latest"` to `.mise.toml`
- [ ] Add `[project]` section with `dependencies` to `pyproject.toml`
- [ ] Update `install-deps.sh` Step 4: replace `pip install` with `uv sync`
[...]

PR #371, opened 2026-06-06
Add ADR-029 for dual Python package management

shrey-bagga:
## Summary

Adds ADR-029 documenting the package-management decision for issue #159:

- keep `requirements.txt` as the shared dependency input for this decision
- require `uv` in the devcontainer bootstrap path
- keep both `pip` and `uv` as explicit non-container contributor choices
- require install, verification, and `language: system` pre-commit hooks to use the selected Python environment
- update the ADR index with the new draft ADR

## Context

This is an ADR-only PR before implementation. It records the architectural direction for issue #159 without migrating scripts, CI, or dependency manifests yet.

## Validation

- Ran a markdown sanity check for trailing newlines, tabs, and conflict markers in `docs/adr/README.md` and `docs/adr/029-dual-python-package-management.md`.
- Confirmed the branch is based on `origin/main` so upstream ADR-028 is preserved and ADR-029 is the next ADR number.

Refs #159

### DIFF
--- a/docs/adr/029-dual-python-package-management.md
+++ b/docs/adr/029-dual-python-package-management.md
@@ -37,11 +37,11 @@ Python development dependency pins remain in [`requirements.txt`](../../requirem
 - `pip` mode installs with `python3 -m pip install -r requirements.txt`.
 - `uv` mode creates or reuses a repo-local `.venv` and syncs the same input with `uv`'s pip-compatible interface, for example `uv venv` followed by `uv pip sync requirements.txt`.
 
-This ADR does **not** migrate dependencies into `[project.dependencies]`, introduce `uv.lock`, or remove `requirements.txt`. Those are packaging and lockfile decisions, not prerequisites for fixing issue #159. A future change may revisit them, but it must update or supersede this ADR rather than folding that migration into the bootstrap fix.
+This ADR does **not** migrate the dependency declaration into `pyproject.toml` (`[dependency-groups]` or `[project.dependencies]`), introduce `uv.lock`, or remove `requirements.txt`. These are independent of the bootstrap fix: `requirements.txt` is already consumable by both `pip` (`-r`) and `uv` (`uv pip sync` / `uv pip install -r`), so no declaration-format change is a prerequisite for fixing issue #159. Deferring the move also keeps this PR off the current six-workflow CI cascade a format change would trigger (see Consequences). A future change may adopt `pyproject.toml` `[dependency-groups]`, but it must update or supersede this ADR rather than fold that migration into the bootstrap fix.
 
 ### D2. Use `uv` unconditionally in the devcontainer
 
-The devcontainer path uses `uv` for Python package installation and package verification. `uv` is declared as a `mise`-managed tool in [`.mise.toml`](../../.mise.toml), and [`.devcontainer/devcontainer.json`](../../.devcontainer/devcontainer.json) invokes:
+The devcontainer path uses `uv` for Python package installation and package verification. `uv` is declared as a **pinned** `mise`-managed tool in [`.mise.toml`](../../.mise.toml) - an explicit version, not `latest` - so the forced-`uv` path stays deterministic (upgrade cadence reconciled with ADR-023 when it lands). [`.devcontainer/devcontainer.json`](../../.devcontainer/devcontainer.json) invokes:
 
 ```bash
 bash scripts/tools/install-deps.sh --python-package-manager uv
@@ -60,17 +60,15 @@ Outside the devcontainer, `install-deps.sh` and `verify-deps.sh` support:
 ./scripts/tools/verify-deps.sh --python-package-manager uv
 ```
 
-The non-container default remains `pip` to preserve the existing contributor path. Contributors opt into `uv` explicitly. `install-deps.sh` passes the selected mode through to its final verification step so installation and verification inspect the same package location.
+Mode selection follows a fixed precedence in both scripts: explicit `--python-package-manager {pip|uv}` flag, then a `RISKMAP_PYTHON_PACKAGE_MANAGER` environment variable, then the default. **Outside the devcontainer the default is `pip`**, preserving the existing contributor path; the devcontainer always passes the flag explicitly (D2), so the default never applies there. `install-deps.sh` forwards the resolved mode to its final `verify-deps.sh` invocation. For a **standalone** `verify-deps.sh` run, the selected mode is recovered from install-time repo-local state, such as a marker written by `install-deps.sh` and/or `.venv` presence as the `uv` signal, before falling back to the precedence chain. A bare verification therefore inspects the same environment that was installed instead of silently defaulting to `pip` and mis-reporting a populated `uv` `.venv`.
 
 ### D4. Bind install, verify, and hook execution to the selected Python environment
 
-`uv` mode is not complete when `.venv` merely exists. Commands that depend on installed Python packages must run through the selected environment:
+**Invariant (required):** in either mode, `pre-commit` installation, commit-time `language: system` hooks, and package verification all resolve the same Python environment that `install-deps.sh` populated for that mode.
 
-- package verification checks packages from `.venv` in `uv` mode and from the ambient selected `python3` in `pip` mode;
-- pre-commit installation runs through the selected Python environment, not a different interpreter;
-- commit-time local hooks declared with `language: system` continue to resolve the same dependency environment that was verified.
+**Delta vs today.** Today every dependency lives in the global `mise`-managed Python's site-packages, and `language: system` hooks resolve them by running `python3` off the `mise`-shim `PATH` - one environment, no binding needed. `pip` mode keeps this behavior. `uv` mode changes it: dependencies live in a repo-local `.venv` that is **not** on the default `PATH`, so without explicit binding a `language: system` hook running `python3` resolves the dependency-empty global interpreter. The binding requirement therefore bites only in `uv` mode, and is the price of `.venv` isolation; leaving it unaddressed would move the #159 mismatch from install/verify into commit time.
 
-The implementation may satisfy the last point by putting `.venv/bin` first on the hook execution `PATH`, by using environment-aware wrapper entries, or by another repo-local mechanism. The architectural requirement is that `uv` mode does not leave `language: system` hooks resolving an unpopulated ambient Python.
+**Illustrative mechanisms (not prescribed):** put `.venv/bin` first on the hook execution `PATH`, use environment-aware wrapper entries, or another repo-local mechanism. The ADR fixes the invariant, not the mechanism, and preserves ADR-005/ADR-013's `language: system` boundary either way.
 
 ### D5. Keep both modes covered by tests, CI, and documentation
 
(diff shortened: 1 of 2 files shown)


davidlabianca (changes requested):
@shrey-bagga thanks for getting this to a clean ADR-only PR. 

The core call (devcontainer forced-`uv`, non-container keeps a `pip`/`uv` choice, fix the install/verify mismatch as an environment-contract problem) is the right shape, and it correctly preserves the ADR-005/013 `language: system` boundary. 

Six changes before I can LGTM and flip this to Accepted — three are substance (the env binding), three are tightening. Inline diffs below.

| # | Change | Severity | Section |
|---|--------|----------|---------|
| 1 | Add **D6 — environment authority & migration** + name the migration test scenarios | major | new `### D6` + Follow-up |
| 2 | Sharpen **D3** — selector precedence + standalone-`verify-deps.sh` mode recovery | major | D3 |
| 3 | Restructure **D4** — lead with the invariant, make today→uv delta concrete | major (clarity) | D4 |
| 4 | Fix **D1**'s rationale (keep `requirements.txt`, but defend it honestly) | minor | D1 + Alternatives |
| 5 | Name the **GitHub Actions** surface + the forced-`uv` CI-coverage gap | minor | Follow-up |
| 6 | Resolve **`uv` pinning** vs ADR-023 + add a `verify-deps.sh` `uv` check | minor | D2 / Consequences / Follow-up |

---

### 1 — New D6: environment authority & migration (major)

**Why.** D5 lists test coverage, but a test needs a *decided behavior* to assert, and the ADR never decides three things: which environment wins in `uv` mode when ambient site-packages are also populated; what happens to a stale/partial `.venv` on a non-container rebuild; and what a `pip`↔`uv` switch does to the prior environment. This is a behavioral gap first, a test gap second — so it earns its own decision rather than just more test bullets.

```diff
 CI may continue to use `pip` from `requirements.txt`. A future workflow may adopt `uv` for speed or parity with the devcontainer, but the repository must retain automated coverage for the `pip` path as long as `pip` remains a supported contributor choice.
+
+### D6. Environment authority and migration
+
+In `uv` mode the repo-local `.venv` is authoritative: verification and hooks inspect `.venv` and ignore ambient/global site-packages, so a `pip`-populated base environment can neither mask a missing `.venv` nor be mistaken for the selected one. `install-deps.sh` is idempotent across rebuilds and reconciles a stale or partial `.venv` by re-syncing it. Switching modes is supported by re-running `install-deps.sh` with the new selector; the newly selected environment becomes authoritative for subsequent verify and hook runs.

 ## Alternatives Considered
```

…and name the scenarios the test follow-up must cover:

```diff
-- Add or update tests for mode selection, `.venv` creation, mode-aware package verification, devcontainer forced-`uv` behavior, pre-commit installation through the selected environment, and preservation of the `pip` path.
+- Add or update tests for mode selection, `.venv` creation, mode-aware package verification, devcontainer forced-`uv` behavior, pre-commit installation through the selected environment, and preservation of the `pip` path. Per D6, also cover: (1) `uv` selected while the ambient env is `pip`-populated — verify checks `.venv`, not ambient, and ambient packages do not let a missing `.venv` pass; (2) `pip`→`uv` switch in place; (3) `uv`→`pip` switch back; (4) non-container rebuild with a stale/partial `.venv` (idempotent reconcile, no manual cleanup); (5) leftover site-packages / PATH carryover does not let `uv`-mode verify false-pass; (6) devcontainer forced-`uv` rebuild is idempotent.
```

---

### 2 — D3: selector precedence + standalone-verify mode recovery (major)

**Why.** D3's last paragraph covers verify only when `install-deps.sh` calls it as its final step. A contributor who later runs a bare `./scripts/tools/verify-deps.sh` has no stated way to learn the install-time mode — with no signal it defaults to `pip` and reports a perfectly good `uv` `.venv` as "packages missing," which is exactly the #159 false-negative this ADR exists to remove, just relocated. The fix is to (a) state the selector precedence and (b) make standalone verify deterministically mode-aware.

```diff
 ### D3. Make the non-container choice explicit
 ...
-The non-container default remains `pip` to preserve the existing contributor path. Contributors opt into `uv` explicitly. `install-deps.sh` passes the selected mode through to its final verification step so installation and verification inspect the same package location.
+Mode selection follows a fixed precedence in both scripts: explicit `--python-package-manager {pip|uv}` flag, then a `RISKMAP_PYTHON_PACKAGE_MANAGER` environment variable, then the default. **Outside the devcontainer the default is `pip`**, preserving the existing contributor path; the devcontainer always passes the flag explicitly (D2), so the default never applies there. `install-deps.sh` forwards the resolved mode to its final `verify-deps.sh` invocation. For a **standalone** `verify-deps.sh` run, the selected mode is recovered from a repo-local marker written at install time (and/or `.venv` presence as the `uv` signal) before falling back to the precedence chain, so a bare verification inspects the same environment that was installed rather than silently defaulting to `pip` and mis-reporting a populated `uv` `.venv`.
```

(The marker mechanism is an implementation choice; the ADR just needs to commit that standalone verify is mode-aware.)

---

### 3 — D4: lead with the invariant, make the delta concrete (major — clarity)

**Why.** The invariant is in there, but it's buried under a per-mode bullet list and a menu of mechanisms, and the before/after is never stated. Separating (i) the required invariant, (ii) the concrete today→`uv` delta, and (iii) the illustrative mechanisms makes the decision and its impact-vs-today legible — and shows the binding requirement bites *only* in `uv` mode.

```diff
 ### D4. Bind install, verify, and hook execution to the selected Python environment

-`uv` mode is not complete when `.venv` merely exists. Commands that depend on installed Python packages must run through the selected environment:
-
-- package verification checks packages from `.venv` in `uv` mode and from the ambient selected `python3` in `pip` mode;
-- pre-commit installation runs through the selected Python environment, not a different interpreter;
-- commit-time local hooks declared with `language: system` continue to resolve the same dependency environment that was verified.
-
-The implementation may satisfy the last point by putting `.venv/bin` first on the hook execution `PATH`, by using environment-aware wrapper entries, or by another repo-local mechanism. The architectural requirement is that `uv` mode does not leave `language: system` hooks resolving an unpopulated ambient Python.
+**Invariant (required):** in either mode, `pre-commit` installation, commit-time `language: system` hooks, and package verification all resolve the same Python environment that `install-deps.sh` populated for that mode.
+
+**Delta vs today.** Today every dependency lives in the global `mise`-managed Python's site-packages, and `language: system` hooks resolve them by running `python3` off the `mise`-shim `PATH` — one environment, no binding needed. `pip` mode keeps this behavior. `uv` mode changes it: dependencies live in a repo-local `.venv` that is **not** on the default `PATH`, so without explicit binding a `language: system` hook running `python3` resolves the dependency-empty global interpreter. The binding requirement therefore bites only in `uv` mode, and is the price of `.venv` isolation; leaving it unaddressed would move the #159 mismatch from install/verify into commit time.
+
+**Illustrative mechanisms (not prescribed):** put `.venv/bin` first on the hook execution `PATH`, use environment-aware wrapper entries, or another repo-local mechanism. The ADR fixes the invariant, not the mechanism, and preserves ADR-005/ADR-013's `language: system` boundary either way.
```

---

### 4 — D1: keep `requirements.txt`, but fix the rationale (minor)

**Why.** The decision to keep `requirements.txt` is the right one for this PR — but the stated grounds are wrong, and the Alternatives bullet conflates four separable decisions (declaration format / lockfile / installer / removal). The key fact the current text omits: **`requirements.txt` is already consumable by both managers** (`pip -r`, `uv pip sync`/`-r`), so no format change is needed for the dual-path goal, and calling the move a "packaging migration" overstates it. Tighten the justification and split the bullet so the deferral is defended honestly. (This is *not* a move to `pyproject.toml` — see the closing note for the now-vs-later trade.)

```diff
-This ADR does **not** migrate dependencies into `[project.dependencies]`, introduce `uv.lock`, or remove `requirements.txt`. Those are packaging and lockfile decisions, not prerequisites for fixing issue #159. A future change may revisit them, but it must update or supersede this ADR rather than folding that migration into the bootstrap fix.
+This ADR does **not** migrate the dependency declaration into `pyproject.toml` (`[dependency-groups]` or `[project.dependencies]`), introduce `uv.lock`, or remove `requirements.txt`. These are independent of the bootstrap fix: `requirements.txt` is already consumable by both `pip` (`-r`) and `uv` (`uv pip sync` / `uv pip install -r`), so no declaration-format change is a prerequisite for fixing issue #159. Deferring the move also keeps this PR off the seven-workflow CI cascade a format change would trigger (see Consequences). A future change may adopt `pyproject.toml` `[dependency-groups]`, but it must update or supersede this ADR rather than fold that migration into the bootstrap fix.
```

```diff
-- **Replace `pip` entirely with `uv sync`, move dependencies into `pyproject.toml`, add `uv.lock`, and remove `requirements.txt`.** Rejected because it forces every contributor and every dependency-update workflow onto `uv`. It also turns a bootstrap reliability fix into a packaging-format migration for a repository that is not currently modeled as an installable Python package.
+- **Move the dependency declaration into `pyproject.toml` `[dependency-groups]`.** Deferred, not rejected. Both `pip` (≥ 25.1) and `uv` can consume a pyproject group — but both already consume `requirements.txt` today, so the move buys no capability needed for issue #159 while cascading to ~13 CI install steps across seven workflows, the `pip` caches, Dependabot, and a pip ≥ 25.1 floor. Sequenced as its own follow-up ADR to keep this PR a bootstrap fix.
+- **Add `uv.lock` as the committed lockfile.** Deferred — an independent lockfile decision, not required for issue #159.
+- **Make `uv sync` the only installer and remove `pip` support.** Rejected because it forces every contributor and every dependency-update workflow onto `uv`. A full `uv`-only packaging migration is also disproportionate to a bootstrap reliability fix for a repository not modeled as an installable Python package.
```

---

### 5 — Name the GitHub Actions surface + the forced-`uv` coverage gap (minor)

**Why.** Two CI items the Follow-up currently misses, both independent of D1. (a) The Follow-up names no workflow at all, yet #159's task list calls out `validate_python.yml`; under this ADR (`requirements.txt` retained) no workflow edit is *required*, but that should be stated rather than left implied. (b) D2 makes the devcontainer forced-`uv` path mandatory, but nothing in CI actually runs it end-to-end — the devcontainer tests are static/unit under pytest — so the most-deterministic path is the least-exercised. Add a Follow-up bullet capturing both.

```diff
 - If a future change wants `[project.dependencies]`, `uv.lock`, or `uv sync` as the repository-wide dependency source of truth, write a follow-up ADR or explicitly supersede this one.
+- GitHub Actions surface: seven workflows install via `pip install -r requirements.txt` with `cache: 'pip'`. Because `requirements.txt` is retained, no workflow edit is required by this ADR. Two open items: (1) the forced-`uv` devcontainer path (D2) has no end-to-end CI coverage today — the devcontainer suites are static/unit; (2) if a future job exercises `uv` in CI it adds a SHA-pinned `astral-sh/setup-uv` action under ADR-024, since `actions/setup-python` does not supply `uv`. (A future move to `pyproject.toml` would instead rewrite those ~13 install steps, the `pip` caches, and one path filter — tracked with that ADR.)
```

---

### 6 — Resolve `uv` pinning vs ADR-023 + add a `verify-deps.sh` check (minor)

**Why.** #159 proposes `uv = "latest"`, but D2's whole point is a deterministic devcontainer path, and ADR-023's reserved concern is exactly pinning — so "should be reconciled with ADR-023" defers a concrete choice to a doc that doesn't exist yet. Take a position now (pin it). Separately, ADR-003 establishes that every `mise`-managed tool carries a `verify-deps.sh` check with a test; `uv` should too.

```diff
-The devcontainer path uses `uv` for Python package installation and package verification. `uv` is declared as a `mise`-managed tool in [`.mise.toml`](../../.mise.toml), and [`.devcontainer/devcontainer.json`](../../.devcontainer/devcontainer.json) invokes:
+The devcontainer path uses `uv` for Python package installation and package verification. `uv` is declared as a **pinned** `mise`-managed tool in [`.mise.toml`](../../.mise.toml) — an explicit version, not `latest`, so the forced-`uv` path stays deterministic (upgrade cadence reconciled with ADR-023 when it lands). [`.devcontainer/devcontainer.json`](../../.devcontainer/devcontainer.json) invokes:
```

```diff
-- `uv` becomes a new `mise`-managed tool dependency in the devcontainer supply chain. Version pinning and upgrade cadence should be reconciled with the reserved devcontainer dependency-pinning policy in ADR-023.
+- `uv` becomes a new `mise`-managed tool dependency in the devcontainer supply chain. It is pinned to an explicit version in `.mise.toml` (not `latest`) for determinism; upgrade cadence is reconciled with the reserved devcontainer dependency-pinning policy in ADR-023 when that ADR lands.
```

```diff
 - Add or update tests for mode selection, `.venv` creation, ... [see change 1]
+- Add a `verify-deps.sh` presence/version check for `uv` (with a test), per ADR-003's pattern that each `mise`-managed tool carries a verification check.
```

shrey-bagga:
@davidlabianca thanks for the detailed review. I updated ADR-029 in `ae7426f` to address the six requested changes:

- Added D6 for environment authority, rebuild idempotency, and `pip`/`uv` mode switching.
- Sharpened D3 with selector precedence and standalone `verify-deps.sh` mode recovery.
- Reworked D4 around the install/verify/hook environment invariant.
- Tightened D1 and split the dependency-format, lockfile, and `uv`-only alternatives.
- Added follow-up coverage for the GitHub Actions surface and forced-`uv` devcontainer CI gap.
- Updated the `uv` pinning language and added the `verify-deps.sh` `uv` presence/version check follow-up.

One small factual adjustment: I verified the current workflow surface and found six Python-installing workflows with 15 `pip install -r requirements.txt` steps, so I used those counts instead of “seven workflows / ~13 steps.”

davidlabianca (approved):
Rework looks good — all six changes landed faithfully and I'm flipping this to Approve.

- D6 (environment authority & migration) reads clearly and the six migration test scenarios give D5 something concrete to assert against.
- D3 now closes the standalone-`verify-deps.sh` false-negative — the precedence chain plus marker/`.venv` recovery is exactly the shape that keeps the #159 mismatch from relocating into bare verification.
- D4's invariant → today-delta → illustrative-mechanisms structure makes the env binding legible, and correctly shows it bites only in `uv` mode.
- D1, the split alternatives, the `uv` pinning position, and the `verify-deps.sh` `uv` check are all in.

And thanks for the correction on the CI-surface numbers in my comment — you're right, it's **six** workflows and **15** `pip install -r requirements.txt` steps, not the "seven / ~13" I'd written. I confirmed against the tree and updated my own notes.

One small nit, non-blocking, in the deferred-`pyproject.toml` parenthetical (Follow-up, last bullet): it says a format change would rewrite "one path filter," but the trigger-path surface is actually two workflows — `validate_python.yml` (`requirements*.txt`) and `persona-pages.yml` (`requirements.txt`) — plus three `hashFiles('requirements.txt')` cache-key references (`validate_python.yml`, `validation.yml`, `validate-issue-templates.yml`). The headline "15 steps across six workflows" is spot-on; only that sub-detail slightly undercounts the blast radius. Fine to fix in place or leave for the follow-up ADR that actually owns the migration.

Approving as-is.

PR #396, opened 2026-06-30
docs(adr): add ADR-030 — agentic component model

davidlabianca:
This adds **ADR-030 (Draft)**, a decision record that defines the component-model taxonomy for agentic, tool-using AI systems. It is **docs-only**: no schema, code, or corpus (`components.yaml`) changes land here — the ADR records *where* the new and existing components should live so the implementation can follow as a deliberate, reviewable change rather than an ad-hoc one. The model was driven by working through CoSAI Workstream 4's Model Context Protocol (MCP) security guidance, which surfaced a cluster of risks and controls with no component to attach to.

**Decisions recorded:**

- A new **top-level `componentsTools`** tier (a fourth peer of Infrastructure / Model / Application) for the external tools trust domain, split into a tool **control** subcategory and a tool **core** subcategory.
- A new **`componentsIdentity`** grouping under Infrastructure to give identity and authorization-decision components a home.
- Rename the **"Model Deployment"** subcategory title to **"Deployment"** (it now hosts model, tool, and runtime workloads; the subcategory id is unchanged).
- **One network policy-enforcement point per trust boundary** (agent boundary and tool-zone boundary), with direction treated as a property of policy, not a separate node (NIST SP 800-207).
- **Two consent surfaces** (application and agent), justified by a distinct agent risk kind — consent fatigue / habituation — not by structural symmetry.
- A serving/hosting fold into **`componentRuntimeHosting`** (alongside `componentModelServing` and a new `componentToolHosting`).
- A dedicated **tool I/O-handling layer** (`componentToolInputHandling` / `componentToolOutputHandling`) at the tool-server boundary.
- A **reasoning-core / orchestration decouple** and agent-exclusive tool invocation, recorded because they are intentional edge-set rewires.

**Deferred follow-ons (not decided here):** a separate **graphical-representation ADR** (typed edge `kind`, directionality, graph composition), and an **autonomy/workload attribute** for the application-vs-agent distinction.

**Implementation footprint this foreshadows:** a new top-level category is a heavy follow-on — it touches the schema enums, the `ComponentGraph` hardcoded-category handling, `mermaid-styles.yaml`, persona ownership of the Tools tier, regenerated diagrams/tables/SVGs, and several test suites. None of that lands in this PR; the ADR exists so reviewers can sign off on the shape before that work begins.

_note: PR reviews act as the ADR flip to `Accepted`_

### DIFF
--- a/docs/adr/030-agentic-component-model.md
+++ b/docs/adr/030-agentic-component-model.md
@@ -1,0 +1,163 @@
+# ADR-030: Agentic component model — external tools tier and identity grouping
+
+**Status:** Draft
+**Date:** 2026-06-27
+**Authors:** Architect agent, with maintainer review
+
+---
+
+## Context
+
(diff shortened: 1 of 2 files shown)


shrey-bagga (approved):
ADR accepted

PR #402, opened 2026-07-06
docs(adr): add ADR-031 — authoring-time agents and skills (Draft)

davidlabianca:
This adds **ADR-031 (Draft)**, a decision record defining the architecture for **authoring-time agents and skills** — reusable, harness-neutral tooling that helps a contributor produce a conformant Control, Risk, Component, or Persona *before* a PR exists. It complements the existing review surface (`content-reviewer`), which only evaluates content someone else already drafted. It is **docs-only**: no agent, skill, schema, or corpus change lands here — the ADR decides the architecture so implementation can follow as deliberate, reviewable work.

The authoring discipline it productizes (altitude rules, classical grounding, counterfactual recording) currently lives only in maintainer-led review and CoSAI paper authoring, applied by hand each time. This ADR gives it a durable, invocable home.

**Decisions recorded:**

- **D1 — Productize the authoring discipline** as first-class artifacts rather than by-hand judgment; where a rule is already recorded (contributing guides, settled practice), the tooling references the source rather than re-deriving it.
- **D2 — Agent-vs-skill boundary:** a **creator/critic agent pair per content type** (Control, Risk, Component, Persona), plus **four skills** — `classical-lexicon`, `mapping-selection`, `altitude-check`, and a framework-mappings audit skill (`audit-framework-mappings`). No per-framework reference-pack skills.
- **D3 — Classical grounding is omnipresent** (baked into every creator, independently checked by the critic). **D3a** adopts NIST as the *canonical naming default* — a practical choice, not a national-authority claim. **D3b** requires *global legitimacy*: each term recorded as a cross-standard equivalence set, with the global standard taking precedence over any single national source.
- **D4 — Out-of-repo framework-knowledge freshness** split by volatility (stable canonical terms pinned, volatile identifiers live-verified), with the framework-mappings audit skill as its single consumer; reuses ADR-027 machinery rather than reinventing pinning/verification.
- **D5 — Boundaries & sequencing:** `content-reviewer` stays the submission gate; sequence Control → Risk → Component → Persona (Persona gated by a *necessity test*, per ADR-021); canonical homes under `scripts/agents/` and a new `scripts/skills/`, with any per-harness wrapper derived from those canonicals (ADR-006).
- **D6 — Skill format:** the **Agent Skills open standard** (`agentskills.io`), adopted on its merits (open, permissively licensed, cross-harness) — the concrete mechanism behind the harness-neutrality of D5.

**Deferred follow-ons (not decided here):** a separate short **ADR-032** for the read-only, consumer-facing exploration skills (a different audience and eval story than this authoring charter); and the vendor-neutralization / promotion of the tooling into `scripts/skills/`, handled by its own effort.

**Implementation footprint this foreshadows:** the creator/critic agent specs and the four skills under a brand-new `scripts/skills/` canonical directory, their per-harness wrappers, the contributor how-to guides, and the eval surface — none of which lands in this PR. The ADR exists so reviewers can sign off on the shape before that work begins.

**Tracking:** Part of #405  (the ADR-decisions child) under epic #404. 

ADR lands as `Status: Draft`; flip to `Accepted` on sign-off.

### DIFF
--- a/docs/adr/031-authoring-time-agents-and-skills.md
+++ b/docs/adr/031-authoring-time-agents-and-skills.md
@@ -1,0 +1,145 @@
+# ADR-031: Authoring-time agents and skills for Risk Map content
+
+**Status:** Draft
+**Date:** 2026-07-01
+**Authors:** Architect agent, with maintainer review
+
+---
+
+## Context
+
+The repository has a mature *review* surface for Risk Map content and no *authoring* surface. `content-reviewer` (ADR-007) reviews YAML in three modes (`diff`, `full`, `issue`); `issue-response-reviewer` composes it into a maintainer comment; the framework-mapping tooling audits and pins mappings. Every one of these tools evaluates content someone else already drafted. Nothing in `scripts/agents/` — and there is no canonical skill surface at all — helps a contributor *produce* a conformant Control, Risk, Component, or Persona in the first place.
+
+Meanwhile, a large body of authoring discipline has been developed — but it only materializes during maintainer-led authoring and CoSAI paper review, applied by hand each time. That discipline is substantial and hard-won, and falls into three strands:
+
+- **Altitude rules.** The right granularity for an entry: specific enough to instruct a reader where a control or risk applies, general enough not to sprawl into per-instance siblings. For controls, state an objective rather than an implementation and do not restate the risk; for risks, a two-test for whether a candidate is genuinely distinct or should merge into an existing one; for components, an absorb-or-decompose base test.
+- **Generalization and classical grounding.** Name an entry by the role it plays, not by a specific product or protocol (those are variations), and align terminology to established security reference architectures — for access control, the NIST SP 800-162 policy-point model (PEP / PDP / PIP / PAP) — rather than coining new terms. The rule "never invent a term when an established term of art exists" runs through all of it.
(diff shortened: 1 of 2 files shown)


shrey-bagga:
Thanks for writing ADR-031. Two small README/index fixes before merge:

1. The branch needs to be updated with current `main`. ADR-030 has merged, and the ADR README now conflicts where this PR adds ADR-031. Please resolve the index so both ADR-030 and ADR-031 are present in order.

2. The ADR-031 README row currently says `reference-pack`, but D2/D4 explicitly reject reference-pack skills and use `audit-framework-mappings` instead. Could you update the row summary to match the ADR?

davidlabianca:
> Thanks for writing ADR-031. Two small README/index fixes before merge:
> 
> 1. The branch needs to be updated with current `main`. ADR-030 has merged, and the ADR README now conflicts where this PR adds ADR-031. Please resolve the index so both ADR-030 and ADR-031 are present in order.

Done...

> 2. The ADR-031 README row currently says `reference-pack`, but D2/D4 explicitly reject reference-pack skills and use `audit-framework-mappings` instead. Could you update the row summary to match the ADR?

Good catch... Legacy README.md desc content... fixed

shrey-bagga (approved):

Issue #405, opened 2026-07-07
docs(adr): land ADR-031 and ADR-032 (authoring-time + consumer exploration architecture)


davidlabianca:
### Summary

Track the two architecture decisions this epic implements from `Draft` to `Accepted`. No implementation lands here — S0 is the decision layer that gates every build child (S1–S9).

### What lands

- **ADR-031 — Authoring-time agents and skills.** A creator/critic agent pair per content type (Control, Risk, Component, Persona) plus authoring skills (`classical-lexicon`, `mapping-selection`, `altitude-check`, and a framework-mappings audit skill); establishes `scripts/skills/` as the canonical skill home authored to the Agent Skills open standard, with per-harness wrappers derived from the canonical (ADR-006).
- **ADR-032 — Read-only consumer exploration skill surface.** The six `explore-*` skills, a distinct read-only family that inherits ADR-031's `scripts/skills/` home and format.

### Current state

- ADR-031: in review as **PR #402** (`Status: Draft`).
- ADR-032: **Draft** on its branch; PR pending.

### Acceptance

- Both ADRs merged to `main` with `Status: Accepted` and their `docs/adr/README.md` index rows in place.
- No implementation is required or included here; S1 may start once these are `Accepted`.

### Notes

- Lifecycle per the ADR README: `Draft` on first commit → `Accepted` on maintainer sign-off. This issue closes when both ADRs are `Accepted`.

Issue #404, opened 2026-07-07
epic: Implement the authoring-time and consumer-facing agent/skill surfaces (ADR-031, ADR-032)


davidlabianca:
### Summary

Two ADRs decide the architecture for a Risk Map agent/skill layer; this epic tracks building it. **ADR-031** establishes an *authoring* surface — a creator/critic agent pair per content type plus a set of authoring skills — that helps contributors *produce* conformant Control, Risk, Component, and Persona entries before a PR exists, and establishes `scripts/skills/` as the canonical skill home authored to the Agent Skills open standard. **ADR-032** establishes a distinct, read-only *consumer* surface — the six `explore-*` skills — that lets framework consumers navigate and understand the existing corpus.

Both ADRs decide architecture only; they author no agent or skill. This epic is the implementation. Everything lands as **vendor-neutral canonical definitions** (`scripts/agents/*.md`, `scripts/skills/*`) with any per-harness wrapper *derived* from the canonical (ADR-006), and each surface ships its own `evals/`.

### Sub-issue grouping (to be filed as children)

Grouped by work-stream along the dependency chain, mirroring the ADRs' own sequencing (ADR-031 D5: skills before agents, Control → Risk → Component → Persona; ADR-032 is parallelizable once the skill foundation exists).

**Decisions** — the architecture the rest of the epic implements:

- [x] **S0 · Architecture decisions (ADR-031, ADR-032, ADR-033).** 
  - Author, review, and land the ADRs (`Draft` → `Accepted` on sign-off). 
  - ADR-031 (#402) and ADR-032 (#407) are **Accepted**; 
  - ADR-033 — the shared vendor-neutral *shipping & lifecycle* standard both corpora conform to — is delivered as a Draft PR (`feature/neutralization-adr`). 
  - *Gates the epic: every implementation child below builds what these decide.*

**Foundation** — unblocks every skill:

- [ ] **S1 · Skill-surface foundation.** 
  - Stand up `scripts/skills/` as a new canonical tracked directory, adopt the Agent Skills format and pin the spec revision the surface targets (ADR-031 D6), and prove the layout end-to-end with the first skill (recommend `classical-lexicon` — self-contained, no cross-refs). 
  - Establish the format/neutrality conformance check. *Blocks S2, S7.*

**Authoring track** (ADR-031) — skills before agents:

- [ ] **S2 · Authoring skills.** 
[...]

PR #407, opened 2026-07-07
docs(adr): add ADR-032 — read-only consumer exploration skill surface (Draft)

davidlabianca:
This adds **ADR-032 (Draft)**, a short decision record establishing a **read-only, consumer-facing exploration skill surface** for the Risk Map — the six `explore-*` skills that let framework *consumers* (security architects, developers, tech/system architects) navigate and understand the existing corpus. It is the consumer counterpart to ADR-031's *authoring* surface, and a distinct family: different audience, read-only (it never mutates the corpus), and a different eval story. It is **docs-only** — no skill or code lands here; the ADR decides the architecture so implementation can follow.

**Decisions recorded:**

- **D1 — A distinct read-only consumer surface.** Its own skill family (audience / read-only / eval story), **inheriting** ADR-031's `scripts/skills/` canonical home (ADR-031 D5) and Agent Skills format (ADR-031 D6) rather than re-deciding them — which is why it is its own ADR, not folded into ADR-031.
- **D2 — Six focused `explore-*` skills, one per query mode** (not a mega-skill, not a router): `explore-risks-by-activity`, `explore-controls-by-classical`, `explore-exposure`, `explore-persona-self-id`, `explore-framework-coverage`, `explain-entity`. Adaptive output, entity-ids cited, framework terms explained inline.
- **D3 — Self-contained skills with thin-sibling reuse** — read the live corpus + generated tables, but defer to an authoring skill for its specialized domain rather than duplicating (ADR-031 D1's defer-don't-duplicate, across the authoring/consumer boundary).
- **D4 — Read-only boundary and handoffs** — explore vs author (ADR-031 creators) vs gate (`content-reviewer`), three separated roles.
- **D5 — Hybrid product→component resolution** for `explore-exposure` (curated lexicon + live lookup; inferred mappings flagged as inferred).

**Deferred (not decided here):** the framework-content / UX specifics — exact query phrasings and the product→component lexicon contents/ownership — belong in a companion `risk-map/docs/design/` note; and the skills' implementation + `evals/` land as separate work.

**Tracking:** Part of #405  (the ADR-decisions child) under epic #404. 

ADR lands as `Status: Draft`.

### DIFF
--- a/docs/adr/032-consumer-exploration-skills.md
+++ b/docs/adr/032-consumer-exploration-skills.md
@@ -1,0 +1,72 @@
+# ADR-032: Read-only consumer exploration skill surface
+
+**Status:** Draft
+**Date:** 2026-07-06
+**Authors:** Architect agent, with maintainer review
+
+---
+
+## Context
+
+ADR-031 productized an *authoring* surface — creator/critic agents plus authoring skills — for contributors *producing* Risk Map content, and established `scripts/skills/` as the canonical skill home (ADR-031 D5) authored to the Agent Skills open standard (ADR-031 D6). That surface serves people who write the corpus.
+
+A separate need is unmet. Framework *consumers* — security architects, developers new to the framework, technology and system architects who are security-literate but not assumed to know the framework internals — want to *explore and understand* the existing corpus (risks, controls, personas, components), not author it. Nothing serves them today: the review surface gates submissions, and the ADR-031 authoring surface mutates the corpus. Neither offers a low-friction, read-only way in.
+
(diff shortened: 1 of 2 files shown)


shrey-bagga (commented):
All looks good. Just the merge conflicts need to be resolved. Also, once ADR-031 gets merged, there might be another merge conflict

davidlabianca:
I'll rebase once -031 lands... If you LGTM I'll just sequence this behind that merge... 

lmk

shrey-bagga (approved):

PR #410, opened 2026-07-10
ADR-034: Dependency-ordered landing sequence for risk-map corpus changes

davidlabianca:
## ADR-034: Dependency-ordered landing sequence for risk-map corpus changes

Addresses #411 

### Summary 
Adds **ADR-034**, a dependency-ordered landing sequence for any change that adds to or modifies the Risk Map corpus (components, personas, controls, risks).

The corpus is a reference graph (`components ← controls`, `components ← risks`, and the `controls ↔ risks` cycle). Landing a referencing entity before its referent, or one half of a reciprocal pair without the other, leaves an intermediate merge broken. This ADR fixes a single topological landing order so the corpus stays validator-green at every step, and a change uses only the subsequence of layers its content touches.

**What it decides**
- **Layers (topological):** new components → new personas → existing-leaf updates → new controls (+ the new risks they require + reciprocal back-links) → new risks (+ reciprocal back-links) → modifications not driven by a new node → deprecations (a status flip to `deprecated`, never enum deletion).
- **Invariant:** every layer merges independently validator-green.
- **Cycle break / tie-break:** reciprocal control↔risk back-links land in the layer that introduces the new node; co-dependent new nodes land in the earlier of their two layers.
- **D3 / D3a:** new components land with their edges (never edgeless); orphan components and personas are *deliberately allowed* — a leaf may land before the controls/risks that reference it — and that allowance is guarded so a future per-PR coverage check cannot silently break the sequence.

**Scope**
- Tooling/process convention → base `main` (per ADR-002). Records the decision and rationale only.
- A follow-up PR adds the contributor how-to guide (`risk-map/docs/`), the `workflow.md` update, and the D3a validator guard.
- Numbered ADR-034 (033 is claimed by an open PR).
- Lands `Status: Draft`; flip to `Accepted` on sign-off.


shrey-bagga:
Non-blocking comment:
Could we clarify whether Layer 7 describes current behavior or the desired future state? Currently, only personas define a deprecated field; the component, control, and risk schemas do not yet support it.
Since an ADR can be forward-looking, those schema changes do not need to be implemented in this PR. However, the text currently says the corpus already supports deprecation for personas and components and that every layer is validator-green.
I suggest documenting deprecation support for components, controls, and risks as a prerequisite or follow-up

davidlabianca:
> Non-blocking comment: Could we clarify whether Layer 7 describes current behavior or the desired future state? Currently, only personas define a deprecated field; the component, control, and risk schemas do not yet support it. Since an ADR can be forward-looking, those schema changes do not need to be implemented in this PR. However, the text currently says the corpus already supports deprecation for personas and components and that every layer is validator-green. I suggest documenting deprecation support for components, controls, and risks as a prerequisite or follow-up

excellent catch - resolved and updated

shrey-bagga (approved):
LGTM

PR #409, opened 2026-07-09
ADR-033: Vendor-neutral shipping and lifecycle for CoSAI agents and skills (Draft)

davidlabianca:
## ADR-033: Vendor-neutral shipping and lifecycle for CoSAI agents and skills (Draft)

Closes #405 
Addresses #408 

### Summary

Adds **ADR-033**, the forward-only charter for how CoSAI ships agents and skills as vendor-neutral, cloneable artifacts, and flips **ADR-031** and **ADR-032** to `Accepted` now that both have merged.

Two commits:
1. **Adds ADR-033 (Draft)** — the shipping standard + lifecycle.
2. **Flips ADR-031 / ADR-032 → Accepted** post-merge (#402, #407), with README index and ADR-033 cross-refs updated to match.

### What ADR-033 decides

The end-state standard the ADR-031 / ADR-032 corpora conform to — it enumerates no corpus and selects no tool:

- **D1 — Canonical-only.** The shipped artifact is the neutral canonical under `scripts/agents/**` / `scripts/skills/**`; no harness-specific wrapper files are tracked.
- **D2 — Neutrality contract.** A machine-checkable core (denylist + a framework-authority allowlist so MITRE/NIST/OWASP/ISO/EU/STRIDE aren't flagged) plus author-judgment guidance.
- **D3 — Consumer leverage.** Consumers adapt to their own harness; the project ships a small, *exercised* worked-example set (≥2 independent harnesses), not wrappers.
- **D4 — Shipping format.** ADR-006 prose agents; Agent Skills `SKILL.md` (per ADR-031 D6, referenced not restated).
- **D5 — A neutrality check is required** — including validation against the adopted standard's reference oracle + required-field core. Constraints fixed; implementation not selected.
- **D6 — Develop / eval / expand lifecycle.** Portable evals required; expansion happens via amendment or a new conforming ADR.

### Review notes

- **ADR-033 lands as `Draft`** — this PR is for review before it's flipped to `Accepted`.
- Deliberately **forward-only**: it describes the shipping standard, not any migration/cleanup history.
- Consistent with the "an ADR constrains a choice, never selects a tool" principle — no linter/eval-runner is named.
- The consumer-adopts approach is evidence-backed (every major harness now has native `SKILL.md` support); the durable risk it names is *silent* frontmatter incompatibility, which D5/D6 address.

Issue #406, opened 2026-07-07
feat(skills): establish scripts/skills/ canonical surface + first skill + conformance gate


davidlabianca:
### Summary

Stand up `scripts/skills/` as a new canonical, tracked directory for vendor-neutral skill definitions, adopt and pin the Agent Skills format (ADR-031 D6), prove the layout end-to-end with one self-contained skill, and add a conformance check. This sets the pattern every later skill (S2, S7) follows.

### Problem

ADR-031 (D5) notes that `scripts/agents/` already holds canonical agent specs, but **skills have no canonical home** — none exists in the repo today. Before any authoring or explorer skill can land, the surface, its format, and its conformance discipline have to exist and be proven once.

### Proposed shape

1. **Directory + layout.** Create `scripts/skills/`. Each skill is a directory with a `SKILL.md` (Agent Skills format — `name` + `description` required) and optional `scripts/`, `references/`, `assets/`, `evals/`.
2. **Format + pinned revision.** Adopt the Agent Skills open standard (ADR-031 D6) and record the spec revision the surface targets, so external changes are taken deliberately rather than silently.
3. **First skill (proves the pattern).** Land one self-contained skill end-to-end — recommend **`classical-lexicon`** (no cross-skill references): the canonical `SKILL.md` + `evals/`. Per ADR-033 D1 the skill is shipped **canonical-only** — no per-harness wrapper is tracked in-repo; harness adaptation is the consumer's (ADR-033 D3).
4. **Conformance gate.** The machine-checkable core of the ADR-033 **D2a** neutrality contract: a denylist (vendor/product/model/harness names, product-named paths) suppressed by a framework-authority allowlist, plus structural frontmatter rules (ADR-031 D5, ADR-004, ADR-033 D2a/D5), wired into `pre-commit`/CI so later skills stay clean.

### Scope boundary

Only the foundation plus the one proof skill. The remaining authoring skills (S2) and the explorer skills (S7) are separate children that follow the pattern this establishes; they are out of scope here.

### Acceptance

- `scripts/skills/` exists with `classical-lexicon` (`SKILL.md` + `evals/`), tracked **canonical-only** (no per-harness wrapper, ADR-033 D1).
- The targeted Agent Skills spec revision is recorded.
- The conformance gate passes on the new skill and **fails on a planted violation** (a vendor name or product-named path), and is wired into the standard checks.
- A short note documents the layout + format for later skills to follow.

### Effort

[...]

PR #428, opened 2026-07-15
Phase 0b · vendor-neutrality pre-commit linter (ADR-033 D2a/D5) + ADR-033 → Accepted

### LINKED ISSUE #408: docs(adr): lands ADR-033 (vendor-neutral shipping and lifecycle for agents and skills)
## Summary

Track ADR-033 — the shared vendor-neutral shipping & lifecycle standard the ADR-031/032 corpora conform to — from Draft to Accepted. Decision layer; no implementation lands here. ADR-033 surfaced while landing S0's two corpus ADRs: it answers the cross-cutting questions neither one owns, and it constrains the build children (S1–S9) — the S1 conformance gate builds to it — without reopening 031/032.

### What lands (the decision — six D's)

- D1 — Canonical-only. The shipped artifact is the neutral canonical under scripts/agents/** / scripts/skills/**; no per-harness wrapper files are tracked.
- D2 — Neutrality contract. A machine-checkable core (denylist + a framework-authority allowlist so MITRE/NIST/OWASP/ISO/EU/STRIDE aren't flagged) plus author-judgment guidance.
- D3 — Consumer leverage. Consumers adapt to their own harness; the project ships a small, exercised worked-example set (≥2 independent harnesses), not wrappers.
- D4 — Shipping format. ADR-006 prose agents; Agent Skills SKILL.md (per ADR-031 D6, referenced not restated).
- D5 — A neutrality check is required — incl. validation against the adopted standard's reference oracle + required-field core. Constraints fixed; implementation not selected.
- D6 — Develop / eval / expand lifecycle. Portable evals required; expansion via amendment or a new conforming ADR.

Acceptance

- ADR-033 merged to main with Status: Accepted and its docs/adr/README.md index row in place.
- The 031/032 Conforms to: ADR-033 cross-refs are present (in the same PR).
- No implementation is required or included here — D2a/D5 (neutrality gate) and D6 (portable-eval requirement) are the spec the S1 conformance gate builds to.

Notes

- Forward-only decision: it defines the shipping standard, not any migration/cleanup history.
- Holds the "an ADR constrains a choice, never selects a tool" line — no linter or eval-runner is named.

davidlabianca:
## Phase 0b · vendor-neutrality pre-commit linter (ADR-033 D2a/D5) + ADR-033 → Accepted

Closes #408
Addresses #406, #404

### Summary

Adds the pre-commit checker that enforces the machine-checkable core of ADR-033's vendor-neutrality contract (D2a) over the two shipped canonical trees, and — now that ADR-033 (#409) has merged and been reviewed — flips it `Draft → Accepted`. This is the front of the promotion stack: the gate every later phase's artifacts pass.

### What's included

- **`validate_neutrality.py` + `_neutrality_data.py`** — the checker plus its maintained denylist/allowlist data module (the one sanctioned place vendor/product/model tokens appear, kept outside the scan scope so it never self-trips). The denylist covers vendor/product/CLI and model-family names, harness-invocation directions, and config paths; a framework-authority allowlist (MITRE/NIST/OWASP/ISO/EU AI Act/STRIDE + their IDs) is never flagged, via real span-overlap arithmetic rather than a whole-line carve-out. Frontmatter rules are structural (parsed YAML).
- **Scope** is `scripts/agents/**` and `scripts/skills/**` only; adoption material (harness-adaptation examples) is intentionally exempt per the ADR-033 D2/D5 clarification.
- **`.pre-commit-config.yaml`** — registers the fail-closed `validate-neutrality` hook.
- **`architect.md` leakage fix** — the linter's first live run flagged a style rule that itself named harnesses as examples; the parenthetical is dropped, keeping the branch self-consistent (its own tree passes the hook it adds).
- **ADR-033 → Accepted** — Status flip on the ADR plus its README index row.

### Review notes

- Fail-closed on unverifiable input (unparseable frontmatter, non-UTF-8, dangling symlinks), hardened via an adversarial audit and an independent review; full suite green.
- v1 is the pattern/frontmatter check. The D5 requirement to also validate shipped skills against the Agent Skills reference validator is deferred to Phase 1, once `scripts/skills/` exists to validate against.

### Ordering

Follows: — (front of the stack; **open this one as a normal PR — the rest are Drafts**). Precedes: **Phase 1 · classical-lexicon** (`feature/phase1-first-skill`).


shrey-bagga:
Thanks for putting this together. The validator is thoughtfully structured, and the test suite and existing checks are passing. I found three implementation cases that may be worth addressing in this PR, along with one question about the intended scope of the initial CLI-command detection.

### 1. Falsy non-mapping YAML can be accepted as empty frontmatter

The frontmatter loader currently uses:

```python
frontmatter = yaml.safe_load("\\n".join(block_lines)) or {}
```

This converts every falsy YAML value into an empty mapping. For example, frontmatter containing `false`, `0`, or `[]` is treated as `{}` instead of being rejected as a non-mapping value.

That appears to differ from the validator’s fail-closed behavior for malformed or non-mapping frontmatter. It also means some structurally invalid frontmatter can pass validation.

Could we please preserve the parsed value and only default to an empty mapping when the YAML document is actually empty? For example:

```python
frontmatter = yaml.safe_load("\\n".join(block_lines))
if frontmatter is None:
    frontmatter = {}
```

The existing mapping check can then reject `false`, `0`, and `[]`. Could we also add regression tests for these cases?

### 2. Changes to the neutrality data do not trigger the pre-commit check

The validator imports its detection terms and patterns from `_neutrality_data.py`, but the pre-commit hook only triggers for files under `scripts/agents/` or `scripts/skills/`.

As a result, changing only the detection data does not cause the affected agent and skill files to be revalidated. For example:

```console
$ pre-commit run validate-neutrality --files scripts/hooks/precommit/_neutrality_data.py
validate-neutrality...................................(no files to check)Skipped
```

This could allow a change to the central policy data to merge without validating the repository content against the new rules.

Could we please make changes to `_neutrality_data.py` trigger a full scan of the agent and skill corpus? Since `pass_filenames: true` is currently used, simply adding the data file to the hook’s `files` expression would pass that Python file to the validator as an input document. The hook may therefore need a separate invocation or another way to distinguish “policy data changed, scan everything” from the normal changed-file path.

A small structural pre-commit test covering this trigger would also help prevent the hook and validator read-set from drifting later.

### 3. The pre-commit and full-scan paths use different definitions of a text file

There are two related file-selection cases:

- The pre-commit expression matches every changed file below `scripts/agents/` and `scripts/skills/`. That includes binary files in supported skill directories, such as `assets/icon.png`. Because explicit filenames are decoded as UTF-8, a changed binary asset can make the hook fail with an invalid UTF-8 error.
- The validator’s no-argument discovery path scans only `.md`, `.py`, `.yaml`, `.yml`, `.json`, and `.txt`. Textual files with extensions such as `.sh`, `.js`, `.ts`, or `.toml` are therefore not included in a full scan.

This is relevant because the skill layout permits bundled directories such as `scripts/`, `references/`, `assets/`, and `evals/`. Some of those directories can contain either textual source files or binary assets.

Could we please align the explicit-file and discovery paths around a shared text-file policy? Ideally, binary assets would be ignored consistently while textual bundled scripts and configuration files would still be checked. Tests with a binary asset and a textual bundled script would help document the intended behavior.

### Question: intended v1 boundary for CLI-command detection

The current CLI detection appears to recognize tool commands only when they are inside inline backticks. For example, a shell-style instruction such as:

```text
$ claude --resume
```

passes validation, while the backtick form is detected.

I noticed that the implementation documents this as a v1 limitation, so I do not want to assume the broader coverage belongs in this PR. Is that narrower scope intentional for the first version? If so, would it make sense to track context-aware detection of shell prompts and fenced command examples as a follow-up? Otherwise, could we please add representative tests and expand the detection here?

shrey-bagga (changes requested):
Requesting above changes.

davidlabianca:
Thanks @shrey-bagga 


> ### 1. Falsy non-mapping YAML can be accepted as empty frontmatter
> The frontmatter loader currently uses:
> 
> ```python
> frontmatter = yaml.safe_load("\\n".join(block_lines)) or {}
> ```
> 

real gap -> will address

> 
> ### 2. Changes to the neutrality data do not trigger the pre-commit check
> The validator imports its detection terms and patterns from `_neutrality_data.py`, but the pre-commit hook only triggers for files under `scripts/agents/` or `scripts/skills/`.
> 
> As a result, changing only the detection data does not cause the affected agent and skill files to be revalidated. For example:
> 
> ```
> $ pre-commit run validate-neutrality --files scripts/hooks/precommit/_neutrality_data.py
> validate-neutrality...................................(no files to check)Skipped
> ```

will add

> ### 3. The pre-commit and full-scan paths use different definitions of a text file
> There are two related file-selection cases:
> 
> * The pre-commit expression matches every changed file below `scripts/agents/` and `scripts/skills/`. That includes binary files in supported skill directories, such as `assets/icon.png`. Because explicit filenames are decoded as UTF-8, a changed binary asset can make the hook fail with an invalid UTF-8 error.
> * The validator’s no-argument discovery path scans only `.md`, `.py`, `.yaml`, `.yml`, `.json`, and `.txt`. Textual files with extensions such as `.sh`, `.js`, `.ts`, or `.toml` are therefore not included in a full scan.
> 
> This is relevant because the skill layout permits bundled directories such as `scripts/`, `references/`, `assets/`, and `evals/`. Some of those directories can contain either textual source files or binary assets.
> 
> Could we please align the explicit-file and discovery paths around a shared text-file policy? Ideally, binary assets would be ignored consistently while textual bundled scripts and configuration files would still be checked. Tests with a binary asset and a textual bundled script would help document the intended behavior.

I see the discrepancy - will work thru options to resolve the tension

> 
> ### Question: intended v1 boundary for CLI-command detection
> The current CLI detection appears to recognize tool commands only when they are inside inline backticks. For example, a shell-style instruction such as:
> 
> ```
> $ claude --resume
> ```
> 
> passes validation, while the backtick form is detected.
> 
> I noticed that the implementation documents this as a v1 limitation, so I do not want to assume the broader coverage belongs in this PR. Is that narrower scope intentional for the first version? 

Yes - that's the initial scope... we can follow-up in the future as we expand our skills coverage, etc... 

> If so, would it make sense to track context-aware detection of shell prompts and fenced command examples as a follow-up? Otherwise, could we please add representative tests and expand the detection here?

I have this tracked as a backlog - no issue landed, at the moment, but will make sure we get it covered in the future

shrey-bagga (approved):
LGTM, thank you for addressing the changes

PR #431, opened 2026-07-15
Phase 2b · altitude-check authoring skill

### LINKED ISSUE #417: S2b · altitude-check authoring skill
## S2b · altitude-check authoring skill

Sub-issue of #404. The absorb-or-decompose / merge-vs-distinct discipline (ADR-031 D2).

### Scope

- The `altitude-check` skill across the control / risk / component tracks (`SKILL.md` + `evals/`).
- A README roster bullet.

### Done when

- The skill lands, passes the gate, and its eval covers all three tracks.

Delivered by the Phase 2b PR (`feature/phase2-altitude-check`).

davidlabianca:
## Phase 2b · altitude-check authoring skill

Closes #417
Addresses #404

### Summary

Adds the `altitude-check` authoring skill — the absorb-or-decompose / merge-vs-distinct test that keeps a draft control, risk, or component at the right level of abstraction. Covers all three tracks.

### What's included

- `scripts/skills/altitude-check/` — `SKILL.md`, `evals/evals.json`.
- One README roster bullet.

### Ordering

Follows: **Phase 2a · mapping-selection** (`feature/phase2-mapping-selection`). Precedes: **Phase 2c · audit-framework-mappings** (`feature/phase2-audit-framework-mappings`).

### DIFF
--- a/scripts/skills/README.md
+++ b/scripts/skills/README.md
@@ -60,3 +60,7 @@ python3 scripts/hooks/precommit/validate_neutrality.py scripts/skills/<name>/SKI
 - `mapping-selection/` — selects a control's/risk's components, addressed
   risks/controls, and framework mappings, grounded in the corpus and the
   framework applicability rules (ADR-031 D2/D4).
+- `altitude-check/` — checks whether a control/risk/component draft is
+  pitched at the right altitude (granularity), applying the per-type
+  altitude tests and deferring terminology to `classical-lexicon`
+  (ADR-031 D2).
--- a/scripts/skills/altitude-check/SKILL.md
+++ b/scripts/skills/altitude-check/SKILL.md
@@ -1,0 +1,44 @@
+---
+name: altitude-check
+description: "Check whether a CoSAI Risk Map entry is pitched at the right altitude (granularity). For a control: objective-not-implementation, not-a-restated-risk, posture-not-mandate, solved-problem, no-duplication. For a risk: the merge-vs-distinct two-test, threat-not-control-gap, and real-not-hypothetical. For a component: the absorb-or-decompose base test, role-not-product, and reader-instructive. Use when authoring or reviewing a control/risk/component draft, when a draft reads like implementation detail or a restated threat, or to decide whether a candidate should be new, merged, or absorbed into an existing entry."
+---
+
+# Altitude Check
+
+Altitude is the granularity an entry is pitched at. Too low and it becomes implementation detail or an instance that sprawls into siblings; too high and it dissolves the gap it was meant to name. Most authoring defects are altitude defects, so check altitude before wording.
+
+Run the tests for the entry type. The control, risk, and component tests are all active; run the set matching the entry type.
+
+## Control altitude tests
+
+Apply each; report pass or adjust-with-fix.
+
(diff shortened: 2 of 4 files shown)


davidlabianca:
@shrey-bagga I can handle the merges of 2b, 2c, 2d (order dependent,etc) if you review and provide approvals...

shrey-bagga (changes requested):
Thanks for putting this together—the scope looks solid, and the local neutrality and format checks pass. Could you please take another look at C1 in `scripts/skills/altitude-check/SKILL.md`?

I may be reading it too literally, but “both must hold to keep a new component” seems inconsistent with the first question asking whether the candidate can be absorbed into an existing component. Would it be clearer to say that a new component should be kept only when it **cannot** be absorbed and its grain still tells readers where controls or risks attach?

Would you also consider adding a C1 eval for the keep-new path? The current component cases appear to cover absorption/rejection outcomes, so that extra case could help clarify the intended decision rule.

davidlabianca:
> Thanks for putting this together—the scope looks solid, and the local neutrality and format checks pass. Could you please take another look at C1 in `scripts/skills/altitude-check/SKILL.md`?
> 
> I may be reading it too literally, but “both must hold to keep a new component” seems inconsistent with the first question asking whether the candidate can be absorbed into an existing component. Would it be clearer to say that a new component should be kept only when it **cannot** be absorbed and its grain still tells readers where controls or risks attach?
> 
> Would you also consider adding a C1 eval for the keep-new path? The current component cases appear to cover absorption/rejection outcomes, so that extra case could help clarify the intended decision rule.

Thanks — both good catches, and you read C1 correctly.

The wording was genuinely backwards: "both must hold" paired with "(1) can it absorb" reads as licensing a new component when it CAN absorb, which is the opposite of the rule. Fixed to negate condition (1) explicitly, matching how `component-creator.md` already states the same test correctly — that file was the source of truth here, and the packaged skill had drifted from it. While fixing it I also named the condition the third outcome (decompose) actually turns on, which the prior wording never stated at all.

On the eval gap: you're right, and it's a bit worse than the one case you flagged — C1 is a 3-outcome test and only "absorb" was ever exercised (no "new," and no "decompose" either). Added coverage for both missing outcomes (2 new cases, 2 decompose cases).

One thing worth flagging on how I built those: a synthetic "keep as new" case is only true against a snapshot of the corpus, and the corpus is actively growing right now (a parallel workstream is adding ~18 components specifically to fill gaps like the ones a synthetic case would target) — so grading against the live `components.yaml` risked the case going stale for reasons that have nothing to do with a skill regression. Rather than accept that risk, I drafted a small pinned fixture that ships with the skill's eval set instead, and worked out with the maintainer whether that pattern deserved recording somewhere durable — it now has: **ADR-033 Amendment 2026-07-30 (D7)** (still _to be filed_ at reply time) formalizes exactly this (absence-grounded eval expectations — "nothing covers this" verdicts — ship with a pinned fixture; presence-grounded ones can stay on the live corpus).

All 4 new cases were blind-exercised (fresh executors, staged skill copy with no access to the answer key), not just authored — and it caught a real bug in my first draft. Both independent runs read my fixture's original description of the input-handling component ("validates/normalizes input") broadly enough to also cover the injection-detection candidate, landing on absorb where new was intended, since the candidate sits at exactly the locus that description named. Not a skill defect — a fixture-wording gap: it described what the component does without excluding the near-miss it needed to be distinct from. Fixed by adding an explicit "does not perform semantic or adversarial-content analysis" exclusion, re-verified clean with a third run.

Branch updated; ready for another look.

shrey-bagga (approved):
LGTM

PR #432, opened 2026-07-15
Phase 2c · audit-framework-mappings skill

### LINKED ISSUE #418: S2c · audit-framework-mappings skill
## S2c · audit-framework-mappings skill

Sub-issue of #404. The framework-mapping correctness discipline (ADR-031 D2).

### Scope

- The `audit-framework-mappings` skill (`SKILL.md` + `evals/`) — defers format/version patterns entirely to `risk-map/docs/contributing/framework-mappings-style-guide.md` rather than shipping its own reference pack, per ADR-031 D2's "no per-framework reference-pack skills" principle. Owns the mapping-format discipline `mapping-selection` defers to.
- A README roster bullet.

### Done when

- The skill lands, passes the gate, and its eval covers the supported frameworks.

Delivered by the Phase 2c PR (`feature/phase2-audit-framework-mappings`).

davidlabianca:
## Phase 2c · audit-framework-mappings skill

Closes #418
Addresses #404

### Summary

Adds the `audit-framework-mappings` skill — checks that an entry's external-framework mappings are well-formed and correctly scoped across the supported frameworks. It owns the framework-mapping discipline that `mapping-selection` defers to.

### What's included

- `scripts/skills/audit-framework-mappings/` — `SKILL.md`, `evals/evals.json`. No internal `references/` — the skill defers entirely to `risk-map/docs/contributing/framework-mappings-style-guide.md` as the single authoritative source for pattern/version-token conventions (ADR-031 D2's "no per-framework reference-pack skills" principle), rather than shipping a second, independently-driftable copy.
- One README roster bullet.

### Ordering

Follows: **Phase 2b · altitude-check** (`feature/phase2-altitude-check`). Precedes: **Phase 2d · audit-identification-questions** (`feature/phase2-audit-identification-questions`).

### DIFF
--- a/scripts/skills/README.md
+++ b/scripts/skills/README.md
@@ -64,3 +64,7 @@ python3 scripts/hooks/precommit/validate_neutrality.py scripts/skills/<name>/SKI
   pitched at the right altitude (granularity), applying the per-type
   altitude tests and deferring terminology to `classical-lexicon`
   (ADR-031 D2).
+- `audit-framework-mappings/` — audits the framework mappings across
+  risks, controls, and personas against the framework-mappings style
+  guide (version pinning, applicability, selectivity), deferring format
+  to the style guide (ADR-031 D2, ADR-027).
--- a/scripts/skills/README.md
+++ b/scripts/skills/README.md
@@ -60,3 +60,7 @@ python3 scripts/hooks/precommit/validate_neutrality.py scripts/skills/<name>/SKI
 - `mapping-selection/` — selects a control's/risk's components, addressed
   risks/controls, and framework mappings, grounded in the corpus and the
   framework applicability rules (ADR-031 D2/D4).
+- `altitude-check/` — checks whether a control/risk/component draft is
+  pitched at the right altitude (granularity), applying the per-type
+  altitude tests and deferring terminology to `classical-lexicon`
+  (ADR-031 D2).
--- a/scripts/skills/altitude-check/SKILL.md
+++ b/scripts/skills/altitude-check/SKILL.md
@@ -1,0 +1,44 @@
+---
+name: altitude-check
+description: "Check whether a CoSAI Risk Map entry is pitched at the right altitude (granularity). For a control: objective-not-implementation, not-a-restated-risk, posture-not-mandate, solved-problem, no-duplication. For a risk: the merge-vs-distinct two-test, threat-not-control-gap, and real-not-hypothetical. For a component: the absorb-or-decompose base test, role-not-product, and reader-instructive. Use when authoring or reviewing a control/risk/component draft, when a draft reads like implementation detail or a restated threat, or to decide whether a candidate should be new, merged, or absorbed into an existing entry."
+---
+
+# Altitude Check
+
+Altitude is the granularity an entry is pitched at. Too low and it becomes implementation detail or an instance that sprawls into siblings; too high and it dissolves the gap it was meant to name. Most authoring defects are altitude defects, so check altitude before wording.
+
+Run the tests for the entry type. The control, risk, and component tests are all active; run the set matching the entry type.
+
+## Control altitude tests
+
+Apply each; report pass or adjust-with-fix.
+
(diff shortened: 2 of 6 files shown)


davidlabianca:
@shrey-bagga I can handle the merges of 2b, 2c, 2d (order dependent,etc) if you review and provide approvals...

shrey-bagga:
Thanks for putting this together—the neutrality gate, eval JSON checks, and repository tests all pass.

Could you please take another look at a few points before approval?

1. I may have misunderstood ADR-031 D4, but should the live identifier check apply to existing mappings as well as newly proposed ones? If so, would you consider adjusting the audit instructions and adding a well-formed-but-nonexistent identifier eval?
2. Issue #418 calls for eval coverage across the supported frameworks, and I did not see an EU AI Act case. Would you consider adding one?
3. The format section says every mapping carries a version token, while STRIDE is intentionally unversioned. Could that sentence be qualified? It may also be worth relying entirely on the authoritative style guide instead of repeating the current patterns in the skill.

One smaller question: was the `references/` directory mentioned in the issue and PR description intentionally omitted in favor of linking directly to the style guide? If so, perhaps the descriptions could be updated to match.

Thanks again—everything else I checked looks clean.

davidlabianca:
> Thanks for putting this together—the neutrality gate, eval JSON checks, and repository tests all pass.
> 
> Could you please take another look at a few points before approval?
> 
> 1. I may have misunderstood ADR-031 D4, but should the live identifier check apply to existing mappings as well as newly proposed ones? If so, would you consider adjusting the audit instructions and adding a well-formed-but-nonexistent identifier eval?
> 2. Issue [S2c · audit-framework-mappings skill #418](https://github.com/cosai-oasis/secure-ai-tooling/issues/418) calls for eval coverage across the supported frameworks, and I did not see an EU AI Act case. Would you consider adding one?
> 3. The format section says every mapping carries a version token, while STRIDE is intentionally unversioned. Could that sentence be qualified? It may also be worth relying entirely on the authoritative style guide instead of repeating the current patterns in the skill.
> 
> One smaller question: was the `references/` directory mentioned in the issue and PR description intentionally omitted in favor of linking directly to the style guide? If so, perhaps the descriptions could be updated to match.
> 
> Thanks again—everything else I checked looks clean.

Thanks for the thorough pass — all four addressed.

**1. Live identifier check on existing mappings, not just new.** Agreed, and there's a real distinction worth drawing out that your framing pointed at: the routine full-corpus audit (overlap, selectivity, structural checks) and a live web-verification lookup are different operations, and broadening the live check shouldn't mean every whole-corpus run starts live-searching every existing entry. Split "Term and identifier verification" into two explicit items: a structural well-formedness check (no live lookup, runs at any scope including whole-corpus, same as the other structural checks) and a live-verification check (per-entity, triggered by a new mapping or a targeted single-entity audit, never an automatic side effect of the routine sweep). Added the well-formed-but-nonexistent eval case you suggested, framed as a targeted audit so it exercises the right path.

**2. EU AI Act eval gap.** Added — a missing-version-pin case in the same style as the existing ones.

**3. STRIDE contradiction / style guide reliance.** Went further than qualifying the sentence — removed the inline pattern snapshot entirely, full DRY. There are 4 open issues right now asking to register new frameworks (ETSI, eIDAS 2.0, SP 800-53/OpenCRE, ISO 42001) — each would've meant touching this list again, so it's not worth keeping even in "as of the current guide" form. The format section now cites the style guide exclusively, with the STRIDE-unversioned exception folded into the opening sentence instead of a separate bullet that could contradict it.

**4. `references/` directory.** You read it right — the skill was always meant to defer entirely to the style guide, never ship its own reference pack (that's the ADR-031 D2 "no per-framework reference-pack skills" principle this skill exists to embody). The issue/PR descriptions were just stale from an earlier plan. Corrected text is ready to apply to both.

One more thing in the spirit of not just claiming this works: the two new eval cases were actually exercised (two independent blind runs, not a read-through), and the exercise caught a real design flaw in my first draft of the identifier-currency case — the identifier I'd chosen tripped a structural crossover check in addition to the live-verification check I meant to isolate, so a future run could pass it without ever doing the live check. Fixed and re-verified clean with a third run.

Branch updated.

shrey-bagga (approved):
Thank you so much for addressing the same. Approved

Issue #343, opened 2026-05-22
conformance: migrate framework-mapping values to the ADR-027 version-pinned form + activate strict schema patterns


davidlabianca:
## Problem

ADR-027 (accepted via #346) makes every framework-mapping value carry a version-distinct, pinned form, and replaces the loose catch-all for the non-ID frameworks with controlled-vocabulary enums. The content in `risk-map/yaml/**` (risks, controls, personas `mappings:`) still holds the legacy, unpinned forms. The schema's strict patterns are defined but dormant (the catch-all is retained) until content conforms. So: **decided + tooled, not yet migrated, not yet enforced.**

## Migration target (per ADR-027 — generated by the #347 tool, not hand-edited)

| Framework | Legacy in content | ADR-027 pinned form |
|---|---|---|
| mitre-atlas | `AML.T0020` (already canonical) | `AML.T0020@<version>` (token-append only) |
| nist-ai-rmf | `GV-6.2`, `MS-2.3`, `MP-…`, `MG-…` | `GOVERN-6.2@<version>`, `MEASURE-2.3@…`, `MAP-…`, `MANAGE-…` |
| owasp-top10-llm | `LLM06` | `LLM06:2024` (`:`-delimited year token) |
| stride | kebab (`information-disclosure`) | PascalCase enum (`InformationDisclosure`) — unversioned, no token |
| iso-22989 (personas) | bare strings (`AI Producer`, `AI Partner (data supplier)`, …) | controlled-vocab enum, pinned (`AI Producer@2022`, …) |
| eu-ai-act | (no content today — prospective, arrives via #164) | `Article N@<version>` when authored |

## Blocked by
- **#347** — implements the ADR-027 schema (version-anchored patterns, controlled-vocab enums, registry fields) + the `framework-mapping-maintainer` generator + validators. **This migration runs the #347 tool to produce pinned values — #347 must land first (generator before content fixup).**
- The decision trio #303 / #318 / #319 (resolved by ADR-027 / #346).

## Work
1. **Run the #347 `framework-mapping-maintainer` tool** to migrate every framework-mapping value in `risk-map/yaml/**` to the ADR-027 pinned form (content → `develop`). Scope: **96 framework-sub-blocks comprising 146 individual values** (risks 100, controls 40, personas 6).
[...]

PR #452, opened 2026-07-27
ADR-030 D1+D2 — external tools category and identity subcategory

### LINKED ISSUE #460: [schema]: Land ADR-030 D1+D2: componentsExternalTools category and componentsIdentity subcategory
Part of #456.

Adds the fourth top-level component category for the external trust domain, an identity
subcategory under Infrastructure, and the closed schema enums that go with them. Structure
only — no components are authored here, which is what lets it merge ahead of the component
layer without leaving the corpus inconsistent.

Implemented in #452.

**Done when:** the category and both subcategories exist in `components.schema.json`'s enums
and the `categories:` block, the `allOf` branches permit exactly the intended
category/subcategory pairs, and `validate_riskmap.py --force --block` is green.

davidlabianca:
## ADR-030 D1+D2 — external tools category and identity subcategory

- target: `develop`
- follow-up for #396
- Closes #460 

### Summary

Implements [ADR-030](docs/adr/030-agentic-component-model.md) decisions D1 and D2, the structural half of the agentic component model. Schema and category structure only — no new components are authored here.

**D1 — external tools category.** Adds the external trust domain as a fourth top-level category, `componentsExternalTools`, with two subcategories, `componentsToolNetworkControls` and `componentsToolInvocationPath`, and recategorizes `componentTools` onto the invocation path.

**D2 — identity subcategory.** Adds `componentsIdentity` under Infrastructure.

### Commits

| Commit | Change |
|---|---|
| 1c457ef | The D1 and D2 taxonomy — schema, corpus, styling, regenerated artifacts |
| 9972926 | Review findings applied |

### Review findings addressed in the second commit

**The category guard could not fail in CI.** `_get_schema_categories()` resolved the schema relative to its own module path. CI copies the validator package to the repository root, so that path climbed above the checkout, the read failed, the error was swallowed, and the guard reported success while checking nothing. It now resolves relative to the working directory, consistent with every other input, and an unreadable schema is a loud failure rather than an empty result.

**The style check passed when styling was missing entirely.** `MermaidConfigLoader` substitutes hardcoded emergency defaults when `mermaid-styles.yaml` is missing or malformed, so deleting the file made every category appear styled. The guard now rejects fallback-served styling. Graph *rendering* still degrades gracefully without the file — only the guard treats fallback as a failure.

**The persona-ownership check was withdrawn.** ADR-030's Consequences required every category to have a persona owner, citing ADR-021. ADR-021 does not establish that requirement; it explicitly records that personas do not participate in per-category enums and that the asymmetry is intentional. The check is removed, the remaining guard is `check_category_style_coverage`, and an erratum on ADR-030 records the correction. Whether categories should carry named owners is a real question, but it is undecided design work rather than a settled rule.

**The taxonomy ids were renamed, twice.** ADR-030 asked for a naming pass before the corpus landed — `componentTools` and `componentsTools` differ by one character. The first pass made each id the camelCase of its title, which also fixed a rendering contradiction: nested subgraph labels derive from the id rather than the title, so a subcategory could render under a name its own `title` contradicted.

The second pass corrected the axis. `componentsToolControlPlane` / `componentsToolDataPlane` split the category by control plane versus data plane, and that split is misapplied here: every component in this category is in-flow, so none of them is a control plane — the actual control plane, the policy decision point and the identity provider, sits in `componentsInfrastructure`. The honest axis is layer, so the subcategories are now `componentsToolNetworkControls` (connection-layer enforcement, wrapping the invocation path) and `componentsToolInvocationPath` (the request path itself, and the primitives the tool server exposes along it). The leaf component `componentTools` is deliberately unchanged throughout.

### Two things a reviewer may want flagged in advance

**Two subcategories land declared but empty.** `componentsIdentity` and `componentsToolNetworkControls` contain no components; the invocation path contains only `componentTools`. This is what D2 specifies — *"neither in the corpus today… This ADR fixes their home; the components themselves are authored through the content process."* The empty subcategories are intentional structure, not an omission.

**One label is unproven by output.** `componentsToolNetworkControls` has no member components, so no subgraph is emitted for it and its rendered label cannot be demonstrated today. It takes the same id-derivation path as the invocation-path subcategory, whose label is verified.

### Verification

- `pytest` — 3566 passed, 6 skipped (3545 at the first commit)
- `pre-commit run --all-files` — all hooks pass
- `validate_riskmap.py --force --block` — edges, lifecycle order, controls/components mirror, category nesting, category style coverage
- `validate_control_risk_references.py --force` — references consistent

Both commits were verified green independently.


davidlabianca:
@shrey-bagga this is the base for the upcoming components expansion based upon the decomposition of the MCP paper

Lands in `develop`

shrey-bagga:
**Review status:** ITERATE

### Summary

The D1/D2 taxonomy, schema, generated artifacts, and issue-template updates are internally consistent, and this content update correctly targets `develop`. The focused tests and current GitHub checks are green. I found one enforcement gap that allows the new category-style guard to be skipped, plus one localized terminology inconsistency in the ADR and corpus description.

### Required changes

#### [major] The category-style guard does not currently block CI or run for all of its inputs

- **Location:** `.github/workflows/validation.yml:195`; `.pre-commit-config.yaml:208-213`; `scripts/hooks/validate_riskmap.py:351-388`
- **Problem:** The new check deliberately reads both `risk-map/schemas/components.schema.json` and `risk-map/yaml/mermaid-styles.yaml`. It only promotes style warnings to a non-zero exit when `--block` is supplied. The pre-commit hook does use `--block`, but its `files:` pattern only triggers for `components.yaml`, `controls.yaml`, and `risks.yaml`. Conversely, the CI component-validation job runs on every PR but invokes `validate_riskmap.py --force` without `--block`.
- **Evidence:** I added an otherwise-valid category to the component schema in an isolated copy of this PR head without adding a style. Both `components.yaml` and `mermaid-styles.yaml` still passed their JSON Schema validations. The current CI command, `validate_riskmap.py --force`, reported the missing style as a warning and exited 0; the same command with `--block` exited 1. This is the exact schema-only case described in `validate_riskmap.py:351-358`, where a category can exist in the schema without appearing in the corpus.
- **Why it matters:** ADR-030 states that a real-corpus guard should fail CI for a styleless category. The repository’s code-review guidance also requires a pre-commit trigger to cover every file a validator reads. As currently wired, a schema-only category change can pass the schema jobs, skip the blocking pre-commit check, and leave the CI check green despite having no styling.
- **Requested change:** Could we add `--block` to the CI component-validation command and expand `validate-component-edges.files` to include `components.schema.json` and `mermaid-styles.yaml`?

#### [minor] The old control-plane/data-plane wording remains after the taxonomy axis was corrected

- **Location:** `docs/adr/030-agentic-component-model.md:33-34,148`; `risk-map/yaml/components.yaml:164-180`
- **Problem:** D1 still labels `componentsToolNetworkControls` as “the tool control plane” and `componentsToolInvocationPath` as “the tool data plane.” The YAML category description likewise says it contains “control-plane components.” The erratum at ADR line 148 says that split is misapplied: every component in this category is in-flow, while the actual control plane—the policy decision point and identity provider—resides under Infrastructure.
- **Why it matters:** ADR-030 is the authoritative explanation of this taxonomy. Keeping the superseded axis in D1 and the source YAML gives future component authors two conflicting descriptions of where authorization and network-enforcement components belong.
- **Requested change:** Could we update D1’s parenthetical descriptions and the YAML category description to use the corrected layer-based framing: connection-layer enforcement around the invocation path, and the invocation path itself?

### Suggested code fix

```diff
diff --git a/.github/workflows/validation.yml b/.github/workflows/validation.yml
@@
-        if python3 validate_riskmap.py --force; then
+        if python3 validate_riskmap.py --force --block; then
           echo "✅ Component edge validation completed successfully"

diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
@@
       - id: validate-component-edges
         name: 'risk-map: component edge consistency'
         language: system
         entry: python3 scripts/hooks/validate_riskmap.py --block
-        files: ^risk-map/yaml/(components|controls|risks)\.yaml$
+        files: ^(risk-map/yaml/(components|controls|risks|mermaid-styles)\.yaml|risk-map/schemas/components\.schema\.json)$
         pass_filenames: false

diff --git a/docs/adr/030-agentic-component-model.md b/docs/adr/030-agentic-component-model.md
@@
-- **`componentsToolNetworkControls`** (the tool control plane) — `componentToolNetworkPolicyEnforcementPoint`, `componentAuthorizationPolicyEnforcementPoint`, `componentFederationProxy`, `componentExternalPromptTemplate`.
-- **`componentsToolInvocationPath`** (the tool data plane) — `componentToolServer`, `componentTools`, `componentToolInputHandling`, `componentToolOutputHandling`.
+- **`componentsToolNetworkControls`** (connection-layer enforcement wrapping the invocation path) — `componentToolNetworkPolicyEnforcementPoint`, `componentAuthorizationPolicyEnforcementPoint`, `componentFederationProxy`, `componentExternalPromptTemplate`.
+- **`componentsToolInvocationPath`** (the request path and the primitives exposed along it) — `componentToolServer`, `componentTools`, `componentToolInputHandling`, `componentToolOutputHandling`.

diff --git a/risk-map/yaml/components.yaml b/risk-map/yaml/components.yaml
@@
-        and application tiers. This category collects the tool and tool-authorization
-        components: the runtime tool services agents invoke, and the control-plane
-        components that mediate and authorize that access.
+        and application tiers. This category collects the runtime tool services
+        agents invoke and the in-flow components that mediate, authorize, and
+        carry those invocations.
```

### Checks run

- Reviewed head: `68b77ad899e31d7bcaafad1f649862d6883b96dc`
- GitHub checks: all reported validation, schema, graph, template, lint, test, build, and CLA checks passed
- Focused changed-area pytest suite: 353 passed
- `python3 scripts/hooks/validate_riskmap.py --force --block`: passed on the unmodified PR head
- `python3 scripts/hooks/validate_issue_templates.py --force`: passed
- `python3 scripts/hooks/validate_control_risk_references.py --force`: passed
- `git diff --check 6fe3fab..68b77ad`: passed
- Schema-only enforcement probe: both JSON Schema validations passed; `validate_riskmap.py --force` exited 0 with a warning; adding `--block` exited 1
- `generate_issue_templates.py --validate`: not runnable from the isolated archive because it requires a Git worktree; the focused template tests and GitHub Template Drift check passed

davidlabianca:
Thanks @shrey-bagga — both findings are right, and the major one is sharper than the write-up claims. Fixed in `341b073`, except for one half that is already decided elsewhere and should not land twice.

## Trigger coverage — taken, and it is an ADR-005 violation rather than a hardening suggestion

You framed the `files:` gap as one of two halves. It stands on its own: the style check reads `components.schema.json` (the category enum, via `_get_schema_categories`) and `mermaid-styles.yaml` (the styles compared against it, via `MermaidConfigLoader`), and neither was in the trigger.

That is ADR-005's Addendum 2026-05-08 invariant:

> **trigger-set ⊇ check-input-set, AND read/discovery-set ⊇ check-input-set.**

with **check-input set** defined as *"for each declared check inside the validator, the file paths whose contents could change the check's verdict"* — which is exactly what both files are here. The invariant is scoped to `pass_filenames: false` hooks, which this one is, and none of the three carve-outs applies: not a generator, not the master-schema fan-out, and its read set is fixed path constants rather than runtime-discovered.

Confirmed before fixing:

```
$ pre-commit run validate-component-edges --files risk-map/schemas/components.schema.json
risk-map: component edge consistency.................(no files to check)Skipped
$ pre-commit run validate-component-edges --files risk-map/yaml/mermaid-styles.yaml
risk-map: component edge consistency.................(no files to check)Skipped
```

Both now run. The trigger uses the two-source alternation your Addendum 2026-06-08 established for the mapping validators — same topology, and this is your own precedent from #343.

**A question while we are here:** that 2026-06-08 addendum — the comparison-oracle one — still reads `Status: Draft (maintainer to flip to Accepted)`. Its Enforcement section is written and this PR is a second motivating instance in a different subsystem, so it looks like an omission rather than a deliberate hold. Is it ready to flip? I did not want to assume, since it is yours.

**Worth flagging separately: the structural test for this was passing vacuously.** `TestTriggerCoverageInvariant` asserts trigger ⊇ *declared* coverage, and the declaration in `_LOCAL_VALIDATOR_TRIGGER_COVERAGE` is maintained by hand — so an oracle nobody registered is invisible to the check that exists to catch it. Registering both makes it fail correctly on the old pattern. The manual-registration design is a broader weakness; filed separately rather than reworked here.

## Wording — taken

D1's parentheticals and the corpus category description now follow the erratum twelve lines above them: connection-layer enforcement wrapping the invocation path, and the invocation path itself. You were right that leaving them gives component authors two conflicting descriptions in one file.

## `--block` in CI — same fix, already decided in #470

This one I would rather not land here, because it is already an accepted decision with reasoning attached. **ADR-037** (PR #470) is the record: D1 sets the rule — every validator pre-commit invokes with `--block` has a CI invocation that also passes `--block`, and where the two differ CI is stricter — and **D2** is your exact change to this job. D1's table lists `validate-component-edges` / `validate_riskmap.py` with "CI state before this decision: invoked without `--block`", "Required: add the flag".

Your probe and that ADR found the same gap independently, which is a good sign for both.

The implementation exists and is green — `chore/ci-block-parity-impl` on the fork — and covers more than this job: `validate_framework_references.py` had the same gap (D4), and three validators pre-commit runs with `--block` had **no CI invocation at all** (D7), which is the failure mode D1 exists to name. It is not filed yet; that is my next step, and it lands on `main` per D6.

**One sequencing risk in the meantime.** #452 targets `develop`, `develop` is currently behind `main`, and the flip lands on `main`. If #452 merges before the next sync, the guard ships **warn-only on the branch that hosts it** — your hole, fully open, in its own home. Filing the implementation is the fix; duplicating the one-line change here would leave the other four gaps open and create a conflict when D2 lands.

## Three things adjacent to your finding, filed rather than folded in

Each is a different mechanism, so none belongs in this PR:

1. **A styled-but-incomplete category still passes every gate, including `--block`.** `mermaid-styles.schema.json` requires only `fill`/`stroke`/`strokeWidth`; `subgroupFill` is optional and the guard tests key presence only. A category with a style missing `subgroupFill` passes schema validation and the guard, then falls through `graphing/base.py`'s substring chain to `#f8f8f8` — subgroups render unstyled, silently. That is the failure this guard exists to prevent, one field over.
2. **The pytest live-corpus guards that would have caught your repro never run for corpus-only PRs.** `validate_python.yml`'s `paths:` is `**.py` plus config files, with no `risk-map/**`. Two guards in `test_mermaid_config_loader.py` catch exactly your schema-only case and are dead for the changes that need them. Working out the right scope for that filter turned out to be its own question — the honest enumeration of what the suite reads is broader than the corpus — so it is being decided separately rather than patched here.
3. **CI's schema-validation loop skips a missing file silently** — `if [[ -f … ]] … else echo "⚠️ Skipping"`. Deleting `mermaid-styles.yaml` produces no schema failure; only `--block` catches it.

Checks after the fix: 3566 passed, hooks green, and the two oracle invocations above now run instead of skipping.

shrey-bagga (approved):
LGTM

PR #454, opened 2026-07-28
fix(agents): content-reviewer.md cites the schema instead of enumerating taxonomy ids

### LINKED ISSUE #453: bug(skills) content-reviewer.md cites the schema instead of enumerating taxonomy ids
### Summary

`scripts/agents/content-reviewer.md` hardcoded the components category list as `componentsData, componentsInfrastructure, componentsModel, componentsApplication`. 

This is wrong today, independent of any branch: `componentsData` is a **subcategory** of `componentsInfrastructure`, not a peer category (confirmed against the live schema enum, which has exactly three: `componentsInfrastructure`, `componentsModel`, `componentsApplication`). 

The risks category list a few lines above had the same shape of problem — currently correct, but a value copy that goes stale the next time the taxonomy changes rather than a citation that stays correct.

Fix is to conform to pattern ADR-031 D1 already mandates ("reference the source rather than re-deriving it").

davidlabianca:
## fix(agents): content-reviewer.md cites the schema instead of enumerating taxonomy ids

Closes: #453 

### Summary

`scripts/agents/content-reviewer.md` hardcoded the components category list as `componentsData, componentsInfrastructure, componentsModel, componentsApplication`. 

This is wrong today, independent of any branch: `componentsData` is a **subcategory** of `componentsInfrastructure`, not a peer category (confirmed against the live schema enum, which has exactly three: `componentsInfrastructure`, `componentsModel`, `componentsApplication`). 

The risks category list a few lines above had the same shape of problem — currently correct, but a value copy that goes stale the next time the taxonomy changes rather than a citation that stays correct.

Both replaced with citations to the schema/corpus, matching the file's own stated rule 16 lines below the bug ("Authoritative enums live in the per-type `*.schema.json` files; consult those rather than inferring") and the pattern ADR-031 D1 already mandates ("reference the source rather than re-deriving it").

### What's included

- `scripts/agents/content-reviewer.md` — the risks and components category bullets now point at `risks.schema.json` / `components.schema.json` + `components.yaml`'s `categories:` block instead of listing values.

### Why now

Surfaced while checking what an in-flight component-taxonomy rename  would break against the agent-spec surface. The rename didn't cause this bug — it's wrong against `main` today — but it's the reason it was found: a citation is timeline-invariant, so this fix has no dependency on and no conflict with that work landing.

### Test plan

- `pytest scripts/hooks/tests/test_validate_neutrality.py` — 152 passed.
- Full suite — 3633 passed, 6 skipped (pre-existing, unrelated).
- Pre-commit hooks pass on the change (vendor-neutrality check: Passed).

### Follow-up

A backlog issue tracks building the automated lint ADR-031 already deferred ("a lightweight consistency check... is deferred to the backlog") so this class of drift is caught mechanically instead of by review.

### DIFF
--- a/scripts/agents/content-reviewer.md
+++ b/scripts/agents/content-reviewer.md
@@ -93,7 +93,7 @@ This agent reads schema and corpus files directly via its own tool access when a
 
 ### ADR Citation Resolution
 
-Corpus content, PR descriptions, and prior review findings sometimes cite a specific decision as `ADR-0NN DN` (e.g. `ADR-030 D5`). Cross-cutting rules are uniformly `D`-numbered across the ADR set — there is no separate `P` tier to disambiguate. Resolve every such citation before relying on it or repeating it in a finding: read `docs/adr/0NN-*.md` and find the `### DN.` heading; do not accept a paraphrase, a title, or an inference about what a decision says as a substitute for its actual text. A rule that a decision states applies "throughout" the document can still sit inside a sub-paragraph whose surrounding heading reads as being about something else entirely — the citation is only as good as the text it resolves to.
+Corpus content, PR descriptions, and prior review findings sometimes cite a specific decision as `ADR-0NN DN` (e.g. `ADR-030 D5`). Most ADRs number cross-cutting rules `D1`, `D2`, ...; some earlier ADRs — e.g. ADR-014 — use `P1`-`P6` instead. Resolve every such citation before relying on it or repeating it in a finding: read `docs/adr/0NN-*.md` and find the heading matching the exact identifier cited (`### D5.`, `### P2.`, whichever the citation names); do not accept a paraphrase, a title, or an inference about what a decision says as a substitute for its actual text. A rule that a decision states applies "throughout" the document can still sit inside a sub-paragraph whose surrounding heading reads as being about something else entirely — the citation is only as good as the text it resolves to.
 
 ### Content Types & Key Fields
 
--- a/scripts/agents/content-reviewer.md
+++ b/scripts/agents/content-reviewer.md
@@ -88,6 +88,8 @@ If provided, use them to supplement (not override) the embedded schema awareness
 
 ## Schema Awareness
 
+This agent reads schema and corpus files directly via its own tool access when a check below says "consult the schema" or "consult the corpus" — that instruction does not depend on the caller pasting schema content into the prompt. The Input Contract's "caller may optionally provide schema files" above describes an inline convenience (skipping a tool call when the caller already has the file open), not a hard dependency the checks below assume.
+
 ### Content Types & Key Fields
 
 **Risks** (`risks.yaml`):
--- a/scripts/agents/content-reviewer.md
+++ b/scripts/agents/content-reviewer.md
@@ -68,6 +68,7 @@ The caller specifies output style via a format flag:
    - `components.yaml`
    - `personas.yaml`
    - `risk-map/yaml/archive/self-assessment-legacy.yaml` (reference only — archived per ADR-021 D6; not a review target)
+   - `docs/adr/` — the tooling/infrastructure ADR set, consulted whenever a corpus claim cites or should be checked against a specific decision (see ADR Citation Resolution below)
 
 **For `issue` mode:**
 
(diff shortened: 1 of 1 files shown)


shrey-bagga:
Thanks—the change points to the correct authoritative sources, and the focused checks pass.

Could you please take another look at the input-contract wording? I may be missing the intended execution model, but the contract currently says schemas are optional and that schema knowledge is embedded, while the new category instructions require consulting those schemas. The composing `issue-response-reviewer` also treats schema files as optional.

Would it make sense to state that the reviewer must have access to the current schemas—either loaded from the cloned repository or supplied by the caller—and describe what it should do when they are unavailable? That would preserve the no-drift benefit without leaving a valid invocation without the source it is expected to consult.

Everything else I checked looks good.

davidlabianca:
> Thanks—the change points to the correct authoritative sources, and the focused checks pass.
> 
> Could you please take another look at the input-contract wording? I may be missing the intended execution model, but the contract currently says schemas are optional and that schema knowledge is embedded, while the new category instructions require consulting those schemas. The composing `issue-response-reviewer` also treats schema files as optional.
> 
> Would it make sense to state that the reviewer must have access to the current schemas—either loaded from the cloned repository or supplied by the caller—and describe what it should do when they are unavailable? That would preserve the no-drift benefit without leaving a valid invocation without the source it is expected to consult.
> 
> Everything else I checked looks good.

Thanks — good catch, and you framed the scope question exactly right.

You're correct that the Input Contract's "schemas are optional, embedded knowledge is primary" language sits oddly under the fixed bullets, which now say to consult the schema directly. It's not actually a logical break: this agent has direct filesystem access, so "consult the schema" means using its own tools, not waiting on the caller to paste one in — the Input Contract's "may optionally provide" language is about skipping a redundant tool call when the caller already has the file open, not a hard dependency the fixed bullets now violate. I've added a short clarifying note right above the Schema Awareness section saying exactly that, so the two don't read as contradictory side by side.

I stopped short of your fuller suggestion (state the reviewer must have access to current schemas, describe fallback behavior when unavailable) — not because it's wrong, but because it's a bigger, pre-existing design question than this PR should carry. The "schema-optional, embedded-knowledge-primary" framing is stated once for the whole file and shared by `issue-response-reviewer.md`, which composes this agent and repeats the same language. Reconciling that properly means rewriting the Input Contract's caller-obligation framing and resyncing the composing agent, not a two-bullet patch — and it's the kind of thing likely to keep resurfacing as more of the Schema Awareness section converts from restated facts to citations (there's a separate backlog item in flight for exactly that sweep). I am filing a backlog issue to track the fuller reconciliation so it doesn't get lost, rather than scope-creeping this fix.

thanks - and ptal

davidlabianca:
@shrey-bagga 

Added one more commit here (0abdbf6) — a small, additive follow-up prompted by the #473 efforts: 

docs/adr/ is now in the Input Contract, and there's a new ADR Citation Resolution note under Schema Awareness. When a review cites ADR-0NN DN, the agent now resolves it against the actual decision text (docs/adr/0NN-*.md, the ### DN. heading) instead of relying on a paraphrase — prompted by a real case where a rule buried in an ADR-030 sub-paragraph got missed and cost a full review cycle downstream.

5 lines, 1 file, purely additive — no changes to the taxonomy-citation fix already reviewed here. Flagging in case it changes the scope of your pass.

shrey-bagga:
Thanks—the schema-access clarification and follow-up issue #459 address my earlier concern.

Could you please take another look at the new ADR Citation Resolution paragraph? I may be reading “across the ADR set” too broadly, but ADR-014 deliberately uses `P1`–`P6`, and the repository contains citations such as `ADR-014 P2`. Would you consider making the resolution rule accept the identifier defined by the cited ADR—such as either `D5` or `P2`—and resolve the matching heading? That would keep valid ADR-014 citations from falling outside the new rule.

davidlabianca:
Thanks @shrey-bagga  — confirmed, and it's on us. The "uniformly D-numbered, no separate P tier" line was written against a retrofit that's queued but hasn't landed yet, not the actual current state of `main`. ADR-014 uses `P1`–`P6` right now, and it's not a rare case — four sibling ADRs cite it that way in their own D1 headings ("Field taxonomy under ADR-014 P2"), so the rule as written would have silently failed on live, load-bearing citations.

Went slightly further than accepting either `D5` or `P2` explicitly: rather than hardcode the two known prefixes (which is itself a small version of the drift problem this whole citation fix exists to prevent — a third tier, or the eventual retrofit, would mean editing the sentence again), the rule now just says to match whichever prefix the target ADR's own heading actually uses, with ADR-014 named as the current example. Self-correcting either way the retrofit question resolves.

Also fixed the same bug in four sibling files you haven't seen yet — the identical wrong instruction got copied to `control-creator.md`, `risk-creator.md`, `component-creator.md`, and `persona-creator.md` when the ADR-citation fix was extended to those agents. Fixed all four the same way rather than waiting for review to reach them.

Branch updated.

shrey-bagga (approved):
Thank you for the changes. Approved

PR #468, opened 2026-07-30
ADR-034 follow-up: D3a guard + contributor landing-sequence guide

davidlabianca:
## ADR-034 follow-up: D3a guard + contributor landing-sequence guide

Implements the follow-up work ADR-034 commissions: the D3a orphan-leaves regression guard, plus the companion contributing guide and its cross-references, so the landing sequence is both enforced and documented for contributors.

Closes #411 
_note: this is the second and final checklist item; the ADR itself landed in #410._

### What this adds

1. **D3a orphan-leaves guard** (`5e81dde`) — a regression test (`scripts/hooks/tests/test_adr034_orphan_leaves_guard.py`) pinning that a component with bidirectional edges but zero control/risk references, and a persona referenced by zero risks/controls, both pass validation today. Pointer-comments in `validator.py` and `validate_control_risk_references.py` flag that coverage is intentionally unchecked, so a future per-PR coverage check can't be added silently without revisiting ADR-034.
2. **Contributor how-to guide** (`2eb1f49`) — new `risk-map/docs/contributing/landing-sequence.md`: the operational companion to ADR-034, covering all 7 layers, the reciprocal-back-link rules (D2a/D2b), the D3/D3a orphan-leaf allowance, and two worked examples (single-layer and multi-layer changes).
3. **Guide alignment** (`2eb1f49`, `8289b3a`) — `workflow.md` gets a pointer for multi-entity-type changes; `guide-components.md`, `guide-controls.md`, `guide-risks.md` each get a one-line cross-reference; `guide-personas.md` is restructured (its old "Adding a Persona" flow bundled persona-wiring into existing risks/controls with the same-PR schema+YAML steps, which contradicted ADR-034's Layer 2 invariant that a new persona lands alone) into an explicit two-PR sequence.

### Scope

- Tooling/process + docs → base `main` (per ADR-002). No `risk-map/yaml/**` or `risk-map/schemas/**` changes — guard test + validator comments + markdown only.
- Depends conceptually on ADR-034, which merged to `main` via #410 on 2026-07-16. The guide and the guard cite it by name rather than by relative link — a holdover from when the ADR was not yet on `main`. Worth adding the link in review or a follow-up; not a blocker.

---

## ADR-034 flipped to Accepted in this PR

[...]

Issue #469, opened 2026-07-30
Enable --block on the CI validation jobs and cover the three ungated validators (ADR-037 D2, D4, D7)


davidlabianca:
### Summary 

Implements ADR-037 D2 and D4.

CI currently runs both `validate_riskmap.py` and `validate_framework_references.py` without
`--block`, while the pre-commit hooks pass it. Local is therefore stricter than the gate that
decides mergeability, and a warn-level regression can pass CI after failing on a
contributor's machine.

### The change

Implements ADR-037 D2, D4 and D7 in `.github/workflows/validation.yml`.

**Strictness (D2, D4)** — two existing invocations gain `--block`:

- the component edge validation step
- the framework reference validation step

**Coverage (D7)** — three validators that CI does not run at all gain invocations, each passing
`--block`, one invocation per validator so a failure names the check that failed:

- `scripts/hooks/precommit/validate_identification_questions.py`
- `scripts/hooks/precommit/validate_yaml_prose_subset.py`
- `scripts/hooks/precommit/validate_prose_references.py`

Two constraints make the coverage half harder than it looks, and both are ways a job goes green
having inspected nothing:

- **Empty file lists (D7a).** These validators take file arguments rather than scanning the corpus.
  Two of the three exit 0 when handed no files. The file set is resolved from each hook's own
  `files:`/`exclude:` regex over the tracked corpus, and the resolver fails loudly rather than
  printing nothing.
- **Path resolution (D7b).** They run **at their real paths**, not copied to the repository root.
  Two derive `--schema-dir` from `Path(__file__).resolve()`, which a copy invalidates.

### Test contract

- **D1 strictness monotonicity** — assert at config level that the flag set CI passes is a superset
  of what pre-commit passes, for every validator supporting `--block`. This is the invariant; the
  line edits are one instance of it, and a future validator added to pre-commit with `--block` but
  not to CI should fail this test rather than recreate the gap.
- **D1 coverage** — assert every pre-commit validator has a CI invocation, so a validator added to
  pre-commit and not to CI fails without anyone editing the test.
- **D3** — assert no workflow step pairs `--block` with a graph-emission invocation.
- **D4** — assert the framework-references job carries the flag, with a non-vacuity check: inject a
  known warn-level violation into a temp corpus copy and confirm the command exits non-zero.
- **D7a** — assert each new invocation's file list is non-empty, and that the resolver fails loudly
  on an unknown hook rather than yielding an empty list.
[...]

PR #470, opened 2026-07-30
docs(adr): land ADR-037 — CI validation authority and --block parity

davidlabianca:
## docs(adr): ADR-037 — CI validation authority and `--block` parity

Related: #469

### The problem

Three validators that pre-commit runs with `--block` had no CI invocation at all. A contributor who has not installed hooks could land a violation with an all-green CI, and nothing in the repository said that was wrong — because no rule quantified over the gap.

### What it decides

**D1 — CI is the enforcing gate; pre-commit is a fast preview of it.** Two parts:

1. **Coverage.** Every validator `.pre-commit-config.yaml` invokes as a **blocking hook** has a CI invocation of at least the same strictness, over at least the inputs the hook would see. A hook is blocking if a violation makes it exit non-zero — whether unconditionally, or under `--block`.
2. **Monotonicity.** Where a validator runs on both surfaces, CI is at least as strict. Never laxer.

Coverage is stated separately because monotonicity alone is satisfied *vacuously* by a validator CI never runs — an empty flag set is trivially a superset of nothing. That is the mechanism by which the three blocking checks came to live only on contributor machines, and a rule quantifying over the intersection of the two surfaces cannot see it.

This settles the developer question directly: contributors who have not installed hooks are fully covered, because nothing depends on their having done so. Hooks move a failure from minutes-after-push to seconds-before-commit. They are an accelerator, not a correctness dependency.

**The quantifier is blocking-ness, not the flag.** `--block` marks a validator that has a warn-only tier to promote; a validator with no such tier blocks unconditionally and never takes the flag. A rule ranging over flag-bearing hooks cannot see those — which is the same defect one surface lower. Six such validators are today invoked by no workflow at all.

The rule resolves to **twelve** hooks. Five are the flag-bearing ones (D2, D4, D7); seven block without a flag (D8), across six validators — one script serves two hooks. The table lists instances, not the rule, and it grew from eight to twelve during review without the rule changing, which is the property it was written to have.

### The decisions

| Decision | Detail |
|---|---|
| **D2** | `--block` added to the risk-map validation job |
| **D3** | `--block` prohibited on graph-generation invocations — the one carve-out, at the level of an *invocation* rather than a validator |
| **D4** | The `validate_framework_references.py` parity gap closes here |
| **D5** | The copy-to-root arrangement stays, with the reasoning recorded |
| **D6** | Landing surface |
| **D7** | The three uncovered `--block` validators gain CI invocations — inputs constructed explicitly (D7a), invoked in place rather than via copy-to-root (D7b), non-vacuity demonstrated per validator (D7c) |
| **D8** | Blocking validators that take no flag are instances, not exceptions — six of them, across seven hooks |

### D8, and why it is separate from D7

D7's three are invoked with `--block` and simply never wired to CI. D8's six are invoked with no flag at all — they have no warn-only tier, so a violation exits non-zero unconditionally — and D1 as first drafted could not reach them.

D8 was drafted naming three validators and widened to six during review. Deriving the set rather than trusting the draft turned up `validate_all_schemas.py`, `validate_persona_site_build.py` and `validate_neutrality.py`, each invoked by a blocking local hook and by no workflow. Naming instances instead of deriving them had reproduced, at the scale of a table, the defect D1 exists to prevent.

Two things the widening surfaced that the draft did not anticipate:

**The fan-out hook needs no carve-out.** `validate-all-yaml-on-master-schema-change` exists to widen scope after its trigger fires, and [ADR-005](../blob/main/docs/adr/005-pre-commit-framework.md) excludes it from the *trigger* invariant for that reason. That exclusion does not transfer to coverage. CI's schema job already validates the same nine pairs, but from a list transcribed into a shell array rather than derived, so a schema added tomorrow is covered by the hook and missed by CI until someone edits the array — the drift D7a already rules on.

**The six are not homogeneous.** `validate-neutrality` declares `pass_filenames: true`, the only file-argument hook among them, so D7a's derive-don't-transcribe requirement reaches it while the other five self-scan. It shares a script with `validate-neutrality-policy`, which self-scans, so CI carries two invocations of one script with different argv shapes. D8 needed no new text for this — "D7a, D7b and D7c apply unchanged" already covers it — but it is worth knowing before reading the implementation.

The gap was not that they were overlooked against the rule; it was that the rule could not see them. Keying the quantifier on a flag rather than on blocking-ness re-introduced, one level up, the exact defect D1 records. D1's coverage clause now ranges over blocking hooks, and these three follow from it rather than being carved in by name.

D7a, D7b and D7c apply to them unchanged. None takes a strictness flag today; if one later grows a warn-only tier and a `--block` to promote it, D1's monotonicity clause governs from that point with no amendment here.

### Addendum 2026-08-04 — D9, and why D1's table stopped being the register

D1 states its rule over blocking-ness: a hook that can fail a commit is governed on arrival and does not land until its CI invocation lands with it. D1's *testable requirement*, two paragraphs later, still scoped both parts to hooks passing `--block`. That sentence predates D8 and was never updated when D8 established that blocking-ness rather than the flag is the quantifier.

The implementation followed the requirement. So the flagged class derives from the config and is genuinely governed on arrival, while the flagless class derives from **this ADR's own instance table** — a validator was governed because someone wrote a row. An adversarial pass confirmed the consequence: a real blocking hook with no flag, no CI invocation and no row lands with a green suite, and deleting a row together with its job and its probe entry is also green.

D9 makes the config the register. Three things made this more than a one-line correction:

- A bare "can fail a commit" predicate resolves to **every hook in the file**. Hooks that fail by writing — formatters, generators — need regenerate-and-diff, not a strictness flag, and no static test distinguishes them. D9a therefore classifies every entry and **fails closed** on one it cannot classify. That is affordable because classification and coverage are separate assertions: the cheap remedy is one registry line, not wiring CI.
- `check-jsonschema` is declared **ten times under one `id`**, so no per-entry id exists for a table row to name. D9a keys on `(id, name)`.
- Four repo hooks that block unconditionally were never rows. All four turn out to have real CI counterparts — including `validate-lifecycle-stage`, whose hook runs `--mode lifecycle` that no workflow runs, but whose check executes in default mode, which CI does run. Measured, not assumed.

D9c keeps the table as a dated record for its before/after columns, mirrored one direction only:
every row must name an entry the config declares. The reverse is deliberately not asserted, because asserting it would make every new validator an ADR edit — exactly what the table's caption disclaims.

**D9 is a decision, not an implementation.** It needs a classification registry, a totality test, the mirror test, and retirement of the table-keyed derivations. It does not block the stacked implementation: D9c leaves the twelve rows unchanged, so the current derivation resolves identically until that work lands.

### Testable requirement

A config-level test parses `.pre-commit-config.yaml` and every workflow, asserting both parts of D1 — that each blocking hook has a matching CI invocation, and that CI's strictness flags are a superset of the hook's. Part one must be written so a validator absent from CI **fails** rather than skips: the intersection-shaped form of that test passes on precisely the gap it exists to catch. Neither part needs subprocess execution.

### Scope

Record only. The implementation lands separately, per D6.


shrey-bagga:
**Review status:** ITERATE

### Summary

The ADR’s technical reasoning is sound and aligns with the current validators, pre-commit configuration, and CI workflows. I found two minor documentation corrections. I also have one non-blocking reminder about ADR landing order.

### Required changes

#### [minor] Correct the unrelated PR reference
- **Location:** PR description, `Closes: #468`
- **Problem:** `#468` resolves to an unrelated merged pull request—the ADR-034 follow-up—not an issue associated with ADR-037. The relevant ADR-037 implementation issue is #469.
- **Why it matters:** The current reference gives readers the wrong change history.
- **Requested change:** Could we remove `Closes: #468` or replace it with `Refs: #469`?

#### [minor] Update the stale D8 count wording
- **Location:** `docs/adr/037-ci-validation-authority-and-block-parity.md:148-154`
- **Problem:** D8 correctly establishes six validators across seven hooks, but the following text still says “these three,” “left three blocking checks,” and “none of the three.”
- **Why it matters:** These counts describe the implementation and test scope, so the prose should agree with the D8 inventory.
- **Requested change:** Could we update those references to distinguish the six validators from their seven hook entries?

### Suggested code fix

```diff
-These three are invoked with no flag at all
+These six validators, across seven hooks, are invoked with no flag at all

-and left three blocking checks living only on contributor machines
+and left seven blocking hooks living only on contributor machines

-these three follow from it
+these six validators across seven hooks follow from it

-None of the three takes a strictness flag today.
+None of the six validators takes a strictness flag today.
```

### Suggestions

- **Non-blocking:** ADR-036 is currently pending in #481 while this PR introduces ADR-037 and links to ADR-036. We should simply be mindful of the landing order so ADR-036 lands before ADR-037 and the numbering and relative links resolve cleanly.

### Checks run
- Reviewed head: `46afdea0a9f13e2fefdf9866b5cf79694cf372e6`
- GitHub primary record: #468 is a pull request, not an issue
- GitHub primary record: #469 is the open ADR-037 implementation issue
- `git diff --check`: passed
- Technical claims cross-checked against the relevant validators, hooks, and workflows
- GitHub checks: CLA passed

davidlabianca:
Thanks @shrey-bagga  — both corrections are in, and the second one was worth catching. Both resolved. ptal


**D8's stale counts** — corrected in `cf443d2`. 

One deviation from your suggested diff, flagged so you can check it: I left the two **`D7's three`**
references alone. D7 still names exactly three validators, so those counts are correct — the staleness
was only in the sentences describing D8. Applying the change to every "three" in that section would
have introduced an error while fixing one.



**On landing order** — agreed, and it's already the plan. ADR-036's record is open as #481 against `main` with no file overlap with this branch, so it can merge first and the relative links here will resolve cleanly. **Approve only - no merge** should be your action when you confirm my changes.  #483 (the implementation) is stacked on this one and should merge third.

shrey-bagga (approved):
Thank you for fixing this. Approved

PR #471, opened 2026-07-30
ADR 030 v2

davidlabianca:
## ADR-030 v2

Redrafts ADR-030 as a v2 that supersedes v1 in place. Closes #461.

> **Two files.** `docs/adr/030-agentic-component-model.md` and `docs/adr/README.md`, +335 / −83. This
> branch no longer sits on another pull request — #452 merged on 2026-08-06 and the branch has been
> rebased onto current `develop`, so *Files changed* now shows this change and nothing else.

> **Landing order.** This merges **before** #473, which lands the corpus implementing it. A reviewer of
> #473 is expected to check components against these decisions — in particular **D15**'s falsification
> gate on `componentAuditRecordRepository` and **D16**'s rule that components land ahead of the
> controls and risks referencing them — so the record needs to be the accepted version first.

### Why a redraft rather than more errata

v1 was a single decision whose implementation diverged from it in five decided places, and the record already carried two errata. At that volume, errata obscure the model rather than clarify it — a reader has to reconstruct the current position from a base document plus a list of corrections. v2 states the model as it actually is.

The convention in `docs/adr/README.md` — status flips to `Superseded by ADR-XXX` when a *different* ADR replaces one — does not apply here. There is no second record: the subject, scope and number are identical.

### Decision numbers are frozen, and the numbering no longer matches reading order

**D1–D10 keep their v1 numbers**, with corrections made in place under them. New decisions append as **D11–D18**.

This is deliberate and it has a cost. The numbering is load-bearing *outside* this document — commit messages, merged pull-request titles, production validators and test suites cite `ADR-030 D1` and `ADR-030 D2` by number, and commit text cannot be rewritten. Renumbering would silently repoint an immutable record at a different decision.

The consequence is that the decisions are numbered in the order they were made, not the order they are best read. **D11 is the load-bearing one** — the model artifact is reachable only by training and by serving — and much of D1–D10 follows from it. The mitigation is a reading table at the head of the Decision section giving the reading path, with the sections themselves left in numeric order so a citation resolves by scanning down.

### What changed relative to v1

**Three places v1 was wrong rather than merely incomplete:**

- a containment edge it required that the corpus deliberately does not carry
- an enumeration of enforcement points that no longer matches the corpus
- a component described as unchanged that had in fact moved

**Two classical attributions, corrected:**

- v1 carried a quotation attributed to NIST SP 800-207 that the publication does not contain.
[...]

PR #473, opened 2026-07-31
[components] Landing the MCP paper's agentic component layer enhancements

### LINKED ISSUE #462: [component]: Land the 17 agentic components from the MCP-security decomposition
Part of #456.

Depends on #452 and the ADR-030 v2 redraft.

Adds the components derived from the CoSAI Workstream 4 MCP-security paper — the identity and authorization policy points, the policy enforcement points, the consent surfaces, the tool-zone data plane, and the hosting and isolation substrate — with their bidirectional edges. Takes the corpus from 25 components to 42.

Carries the updates to existing components in the same change. The additions and the rewiring are one architectural move: the model artifact becomes reachable only through the serving tier, which is why serving relocates and why several existing components lose direct edges. Split apart, the additions would sit in a topology that no longer explains them.

Per [ADR-034](docs/adr/034-corpus-change-landing-sequence.md) this is Layer 1 — components land before the controls and risks that reference them, and the interim absence of control coverage is the sequence working, not a gap.

**Done when:** every new component carries at least one bidirectional edge, no component is isolated, the generated diagrams, tables and SVGs are regenerated, and each commit is independently green.

davidlabianca:
## [components] Landing the MCP paper's agentic component layer enhancements

Lands the agentic component layer from the CoSAI Workstream 4 MCP-security decomposition.

Closes #462

> **Compare against `develop`.** This branch no longer sits on another pull request — #452's schema
> landed in `develop` on 2026-08-06, so what you see here is this change and nothing else: 27 files,
> +2,604 / −686.

> **Landing order.** **ADR-030 v2 (#471) merges first**, then this. Both target `develop`. The
> decision record has to be the accepted version before the corpus implementing it lands — every shape
> decision below cites it, and two of them (D15, D16) are what a reviewer would use to reject a
> component here.

### What lands

**17 net-new components, 0 removed** — the corpus goes from 25 to 42. Two existing components are
recategorized. No component id changes meaning; every prior reference still resolves to what it
referred to.

| Tier | Components |
|---|---|
| Identity (`componentsIdentity`) | identity provider, authorization policy decision point, federation proxy |
| Deployment substrate | runtime hosting, tool hosting, isolation runtime, audit record repository |
| Agent | agent network policy enforcement point, agent-tool transport, agent consent surface |
| Application | application network policy enforcement point, application consent surface |
| External tools | tool server, tool network policy enforcement point, tool input/output handling, authorization policy enforcement point |

Recategorized: `componentModelServing` → `componentsModelCore`, `componentModelStorage` →
`componentsDeployment`. Both follow from ADR-030 D11 and D12 — once the model artifact is reachable
only through serving, serving is a model-tier concern rather than a deployment one.

### The decision record

Every shape decision here is recorded in **ADR-030 v2** (#471), not in this PR. The two rules doing
most of the work:

- **D4 — operator authority individuates enforcement components.** This is why there are three network
  enforcement points rather than one: the agent operator's egress policy, the application operator's
  egress policy, and the tool provider's ingress policy are three different reviewed policies under
  three different authorities. It is also why shared resources stay whole — one audit record
  repository written by many components, not one per writer.
- **D6 — code ownership individuates hosting substrates.** `componentRuntimeHosting` and
  `componentToolHosting` divide on whether the hosted code is the system's own or a third party's,
  because that is what changes the isolation design.

Components are **classes, not instances**. A deployment holds as many instances of one as its
authorities require — the identity provider a tool server demands is a different instance from the
agent's own, and both are the same component.

### The prompt-template fold

MCP defines three primitives a tool server exposes: tools, resources, and prompt templates. An earlier
revision of this branch gave prompt templates their own component. They do not need one, because a
component is earned by a distinct locus rather than by a distinct payload class, and the three
primitives fail differently:

| primitive | poisoned → | where the failure lives |
|---|---|---|
| executable tool | a wrong action | the capability |
| retrievable resource | wrong data | the capability |
| prompt template | the server steering the model | the advertisement |

So `componentTools` is retitled **External Tool Capabilities** and widened to cover both the executable
and the retrievable, and the instruction-shaped template joins `componentToolServer` — the endpoint
that already owns advertisement, discovery, and per-connection capability negotiation, and whose
existing over-advertisement threat already describes a substituted template.

A separate resources component was drafted and rejected on the same test: its two claimed obligations
already had homes. Read-scope authorization is enumerated verbatim by
`componentAuthorizationPolicyEnforcementPoint`, and identifier-to-content binding belongs to the tool
server, since the artifact does not resolve its own identifier.

### Where reviewers should push

- **The three enforcement points.** If D4's individuation rule is too permissive or too strict, this is
  where it shows. The alternative — one enforcement point per traffic direction — is recorded as
  rejected in ADR-030 v2's Alternatives, on the grounds that a second node is earned by a second
  authority, not a second direction.
- **`componentAuditRecordRepository`.** It carries a falsification gate in D15: at least one control
  must state an obligation that could not have been stated on any emitter. If you do not think that
  gate is met, the component does not deserve its node.
- **Zero control coverage is deliberate**, per ADR-030 D16 and ADR-034: components land ahead of the
  controls and risks that reference them, so this PR adds components no control yet points at. That is
  the layering, not an omission — the controls and risks follow in their own change, and that change is
  already drafted against this branch.

### The registry consult

A question ADR-030 v2 D14 left open and then closed: which side owns the tool-registry consult.

It is the **caller's**. `componentToolNetworkPolicyEnforcementPoint` enforces the tool *provider's*
ingress policy, and `componentToolRegistry` is the AI system's own catalog — a remote provider does not
consult the caller's catalog to decide whether to admit the caller. The vocabulary agrees: the
tool-side point enforces allowed *sources*; the registry supplies endpoints the system is permitted
*to reach*, which is an allowed-destination list.

### Commits

| | |
|---|---|
| `2e7b097` | renames the `componentsModelDeployment` subcategory to `componentsDeployment` |
| `f6aee0c` | adds the agentic MCP-decomposition component layer |
| `49387b3` | resolves the registry consult on the caller's side of the boundary |
| `161426f` | applies the review-gate findings across the layer |
| `264666c` | folds prompt templates into the tool server and widens capabilities |
| `f8e1d12` | makes the two singular-reading repairs an earlier commit claimed |


davidlabianca:
@shrey-bagga as discussed we will move this towards `merge to develop` to facilitate reviews

shrey-bagga (approved):
As discussed, this is fine to merge into develop. We will do a proper review with @imolloy and @sarahnovotny and WS4 team before we merge to main

Issue #477, opened 2026-08-01
[infra] Remove the control and risk diagram generators, keeping only --to-graph


davidlabianca:
## [infra] Remove the control and risk diagram generators, keeping only --to-graph`

`validate_riskmap.py` generates three diagrams. Two of them — the control-to-component graph (`--to-controls-graph`) and the control-to-risk graph (`--to-risk-graph`) — are going away for lack of usage and poor usability. The component graph (`--to-graph`) stays.

**Scope fence: diagrams only.** This is not a deprecation of control or risk content, of any table, or of any validator. Naming that up front because the flag names invite over-reach.

### What stays, explicitly

- Every file under `risk-map/tables/`, including `controls-full.md`, `controls-xref-components.md` and `controls-xref-risks.md`. They come from `yaml-to-md` and share nothing with the graph generators.
- The controls↔components mirror check.
- `validate_control_risk_references.py`, which parses independently.
- `--to-graph` and `risk-map-graph.{md,mermaid,svg}`.
- `controls.yaml` and `risks.yaml`.

### What goes

| | |
|---|---|
| CLI | `--to-controls-graph`/`-c`, `--to-risk-graph`/`-r`, their epilog examples and generation blocks |
| Modules | `graphing/controls_graph.py` (451 lines), `graphing/risks_graph.py` (296 lines), and their exports |
| Artifacts | `diagrams/controls-graph.{md,mermaid}`, `diagrams/controls-to-risk-graph.{md,mermaid}`, `svg/controls-graph.svg`, `svg/controls-to-risk-graph.svg` |
| Hooks/CI | two of three forms in `precommit/regenerate_graphs.py`; the graph-comparison loop in `validation.yml`, which is array-driven over three graphs and collapses to one |
| Config | `graphTypes.control` and `graphTypes.risk` in `mermaid-styles.yaml` **and** the matching schema sections — the schema is `additionalProperties: false`, so these move as one unit |
| Tests | `test_risk_graph.py` (29 tests) outright; `test_regenerate_graphs.py` (26 tests) shrinks |
| Docs | 13 files carry the flag names |

### The parts that need judgment, not just deletion

**`ControlGraph` has no dedicated test file.** Its references are spread across five suites, and several use it as a convenient *second* graph type to prove that a `BaseGraph` or config-loader behaviour generalizes rather than being special-cased. Deleting those references would quietly reduce coverage of code that stays. They should be re-pointed at a surviving graph type.

[...]

PR #478, opened 2026-08-01
fix(hooks): validate_riskmap.py --file is parsed and never read

### LINKED ISSUE #476: [bug] validate_riskmap.py --file is never read — documented invocation exits 0 without validating
## Summary

`scripts/hooks/validate_riskmap.py` parses `--file PATH` and never reads it. The documented invocation exits 0 without opening the file.

```
$ python3 scripts/hooks/validate_riskmap.py --file risk-map/yaml/components.yaml
🔍 Checking for staged YAML files...
   No YAML files to validate - skipping
$ echo $?
0
```

That is the epilog's own example, verbatim, run against the real corpus with nothing staged. A path that does not exist behaves identically — same output, same exit 0.

**Cause.** At `validate_riskmap.py:209-214`, file selection ignores `args.file` in both branches:

```python
if args.force:
    target_file = DEFAULT_COMPONENTS_FILE   # --file discarded
else:
    target_file = None                      # --file discarded; falls back to staged files

yaml_files = get_staged_yaml_files(target_file, args.force)
```

`args.file` appears nowhere else in the program.

**Impact.** The flag is advertised in the module docstring, the `--help` epilog, and `risk-map/docs/guide-components.md` step 5, which tells contributors to validate during development. Any single-file check run against a candidate or work-in-progress file has reported green without reading it — a vacuous pass, not a false negative, so nothing in the output distinguishes it from a real one.

**Why the suite missed it.** `test_validate_riskmap.py` has 43 tests and lists `--file` among what it covers. Two touch it:

- `test_parse_args_with_file_path` asserts `args.file == Path(...)` — that tests argparse, and would pass unchanged if every line after `parse_args()` were deleted.
- `test_main_force_mode_uses_default_file` asserts what reaches `get_staged_yaml_files`, which is the right seam, but passes `--force` alone and so pins the current behaviour as correct.

The remaining `main()` tests patch `get_staged_yaml_files` with a fixed return value, stubbing out file selection entirely. No test passes `--file` into `main()`.

**Proposed fix**

1. Honour `--file` in the non-force path so the documented invocation validates the named file.
2. Fail loud on a missing or unreadable path instead of exiting 0.
3. Decide `--force --file` deliberately. The epilog presents the two as alternatives (`--force` = default file, `--file` = specific file), so erroring on the combination is defensible; silently preferring one is not.

Test-first. The RED test constructs what no existing test does — `main()` with `--file` — and asserts the path reaches `get_staged_yaml_files`. `test_main_force_mode_uses_default_file` passes no `--file` and stays green through the fix, so this is additive rather than a rewrite of existing assertions.

**Scope:** `scripts/hooks/validate_riskmap.py` and its test suite. No corpus, schema, or generated-artifact change. Infrastructure, so it branches from and merges to `main`.

davidlabianca:
## fix(hooks): `validate_riskmap.py --file` is parsed and never read

Closes: #476

`validate_riskmap.py` accepts `--file PATH` and does nothing with it. The documented invocation exits 0 without opening the file:

```
$ python3 scripts/hooks/validate_riskmap.py --file risk-map/yaml/components.yaml
🔍 Checking for staged YAML files...
   No YAML files to validate - skipping
$ echo $?
0
```

That is the `--help` epilog's own example, run against the real 42-component corpus. A path that does not exist behaves identically. `args.file` appears nowhere in the program outside its `add_argument` call.

### What was actually broken

Three places discarded the flag, not one:

1. **File selection** never consulted `args.file`.
2. **The validation target** was hardcoded to the default corpus. `yaml_files` was consumed only for its length and truthiness, so the files selected and the file validated were fully decoupled — the `Found N YAML files to validate` line reported work the program did not do.
3. **The downstream cross-corpus checks** read hardcoded repo-relative paths regardless.

Site 2 is deliberate when `--file` is absent, and stays: the hook fires on `controls.yaml` and `risks.yaml` too, so a commit staging only `controls.yaml` must still validate `components.yaml`. The obvious refactor there breaks every controls-only commit, so it is fenced by test.

### What lands

| Commit | |
|---|---|
| `79d09cf` | The parse layer — one error type for unusable input, and two error messages that name the file they were given |
| `c7905bd` | The flag — selection, the rejected combinations, the skip rules, and the characterization suite |

**`--file` now selects the corpus**, short-circuiting selection rather than routing through `get_staged_yaml_files`. That function consults git on the non-force path and discards the result when a target is given, so routing through it would tie the flag to the staging it exists to ignore, and make it exit 0 having validated nothing outside a repository.

**Two combinations are rejected (exit 2) rather than accepted and discarded.** `--force` and `--file` both answer *which corpus*. Lifecycle mode reads a fixed path `--file` cannot name. `--file` is also rejected with `--to-controls-graph` and `--to-risk-graph`: those read the repo's own `controls.yaml` and `risks.yaml`, so joined to a named corpus they render every control with **zero** component edges and exit 0 on a diagram that looks finished. `--to-graph` stays — it is the one graph drawn from the corpus `--file` names. All four constraints are in the flag's help text.

**The repo-scoped checks skip when `--file` names a corpus other than the default, and say so.** Lifecycle order and the controls mirror read other repo files; against a named corpus they report manufactured findings. Category nesting is *not* repo-scoped — it compares each component's pair against the `categories:` block in the same file — so it runs. The comparison is on resolved paths, so naming the default corpus explicitly is not a weaker run than `--force` over the same bytes.

**An unusable corpus exits 2 naming the file and the reason** — on the default path as well as under `--file`. A malformed `components.yaml` previously told a contributor to report a bug against this tool.

### Nothing that ships today changes

This modifies a hook that gates every commit in the repository, so the constraint was that every existing invocation behave identically. Five forms ship: pre-commit `--block` and `--mode lifecycle`; CI `--force` and its three graph forms; and the `regenerate-graphs` hook, which passes no `--force` and writes tracked diagrams.

All five were diffed — stdout, stderr, exit code and generated artifact bytes — against the pre-change script in matched sandboxes, including the CI root-copy layout, genuinely staged indexes, and dirty corpora with and without `--block`. Identical throughout.

`TestProductionInvocations` locks that behaviour: which corpus reaches validation, which downstream checks ran, and the output CI greps. Twelve mutations are each caught by a named test — among them a one-token slip that silently disables the isolated-component gate, and a path normalisation invisible to a test using a `.md` filename when CI passes `$(mktemp)`.

Suite: **3548 passed**.

### Deliberately not here

Three defects in the same parse layer are tracked in #477, which already opens `utils.py`:

- `parse_controls_yaml` and `parse_risks_yaml` carry the same hardcoded-filename defect. Latent — the only call site passing an explicit path passes the literal default and guards on `.exists()` first — and #477 deletes one of the two outright.
- An **empty** `components.yaml` still reports `TypeError: 'NoneType' object is not subscriptable`.
- The corpus is opened twice per run: once to parse, once by the nesting check for the `categories:` block that `parse_components_yaml` does not return. Closing that means changing its contract.

**The last one is pinned rather than noted.** A test asserts the corpus is opened *exactly twice* and cites #477. A fix that drops it to one will fail that test, so whoever lands it has to update the count and close the deferral; a careless third read fails it too.

### Reviewing

The two commits are independently green and split by subject rather than by test-then-code, so `79d09cf` can be read as a self-contained contract change before the flag that uses it.

Worth pushing on: whether rejecting `--force --file` is right rather than composing them (`get_staged_yaml_files` has a branch that would have supported the combination, though its non-force sibling accepts a target identically, so it is not evidence of intent); and whether `--mode lifecycle --file` should be rejected while `--mode lifecycle --to-graph` is still silently ignored. That asymmetry is knowingly unreconciled, and making `--file` name the lifecycle corpus stays open.

### DIFF
--- a/scripts/hooks/riskmap_validator/utils.py
+++ b/scripts/hooks/riskmap_validator/utils.py
@@ -22,7 +22,7 @@ def parse_components_yaml(file_path: Path = None) -> dict[str, ComponentNode]:
         file_path = Path("risk-map/yaml/components.yaml")
 
     if not file_path.exists():
-        raise FileNotFoundError(f"Controls file not found: {file_path}")
+        raise FileNotFoundError(f"Components file not found: {file_path}")
 
     try:
         with open(file_path, "r", encoding="utf-8") as f:
@@ -85,7 +85,7 @@ def parse_components_yaml(file_path: Path = None) -> dict[str, ComponentNode]:
     except yaml.YAMLError as e:
         raise yaml.YAMLError(f"Error parsing components YAML: {e}")
     except KeyError as e:
-        raise KeyError(f"Missing required field in components.yaml: {e}")
+        raise KeyError(f"Missing required field in {file_path}: {e}")
 
 
 def parse_controls_yaml(file_path: Path = None) -> dict[str, ControlNode]:
--- a/scripts/hooks/riskmap_validator/validator.py
+++ b/scripts/hooks/riskmap_validator/validator.py
@@ -20,12 +20,26 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
+import yaml
+
 from .models import ComponentNode, ControlNode
 from .utils import parse_components_yaml
 
 
 class EdgeValidationError(Exception):
     """Custom exception for edge validation failures."""
+
+    pass
+
+
+class CorpusParseError(Exception):
+    """
+    Raised when a file cannot be parsed into a components corpus.
+
+    Single type for every unusable-input shape so callers can report a bad
+    input file without having to enumerate the exception types
+    parse_components_yaml happens to surface.
+    """
 
     pass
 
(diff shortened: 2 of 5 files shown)


shrey-bagga (approved):
**Review status:** APPROVED

### Summary

This correctly wires `--file` through corpus selection and validation while preserving the existing production invocations. Invalid inputs are reported cleanly, and repo-scoped checks are skipped only for genuinely custom corpora. Rejecting `--force --file` and `--mode lifecycle --file` is coherent and avoids silently ignoring user input. I found no correctness, security, or maintainability issues requiring changes.

### Checks run

- Reviewed head: `c7905bdacc3b1d301162a1f680c9c9feb0a69be8`
- `git diff --check`: passed
- Exact-head focused tests: `172 passed`
- Merged with current `main` (`0e523d0`) and related CI parity tests: `400 passed, 9 skipped`
- Ruff on all five changed files: passed
- Full merged suite: `3810 passed, 19 skipped, 3 environment-specific failures`
- The same three failures reproduced on pristine `main`, confirming they are unrelated to this PR
- GitHub Python validation and CLA checks: passed

PR #481, opened 2026-08-02
[adr] File ADR-036 — decoupled component-graph emission

### LINKED ISSUE #472: [adr]: File ADR-036 (decoupled component-graph emission) as a decision record
### The gap

ADR-036 governs the decoupled component-graph emission work and is cited as binding by #464 and #465. It has never been filed — it is not on `main` or `develop`, and it has no pull request.

#464's central constraint is quoted from ADR-036 D7. A reviewer of #464 cannot verify it, because the document is not in the repository.

### What ADR-036 decides

Nine decisions covering the emission architecture:

- **D1** — adopt the topological rewrite; reject layout-engine tuning
- **D2** — two-layer architecture: a transform IR, then an emission pass
- **D3** — the configuration surface and the rollback lever (`mode: flat` / `mode: decoupled`)
- **D4** — aspect membership and aggressiveness: conservative, lift observability only
- **D5** — channel granularity, split by trust concern
- **D6** — channel arm splitting, one arm per distinct target
- **D7** — determinism, stable ids, and fail-loud guards *(the constraint #464 cites)*
- **D8** — display-layer acronym substitution: a closed, declared set, never touching ids
- **D9** — consult-class channel landing: route to the component, not to the gate's `_in` port

### Scope

**The record only.** The document is already separable — it is the first commit on the emission branch, ahead of all six implementation commits, and touches only the ADR and the `docs/adr/README.md` index row.

Base `main`, per ADR-002: this is a tooling decision. Note the deliberate split — the **implementation** (#464, #465) targets `develop`, because it rewrites generated artifacts under `risk-map/`. The decision and its implementation land on different branches for the same reason ADR-030 v2 and its corpus layer do.

Lands `Status: Draft` for maintainer acceptance.

### Why not file now

Before the record PR is cut, ADR-036 needs a reconciliation pass against what ADR-030 v2 and the component layer actually landed. Two defects are already confirmed, and they are the reason rather than a hypothetical:

**1. It names a component that will not exist.** ADR-036 refers to `componentSecureLogging` in three places, including a Consequences bullet that is load-bearing — the one describing which node carries the `aspectStyle` marker and the live "N writes lifted" count in the rendered graph. That component is renamed to `componentAuditRecordRepository` in ADR-030 v2 and on the component branch. As written, ADR-036 describes a node id that is not in the corpus it draws.

**2. It asserts a reciprocal pointer that does not exist.** ADR-036's Follow-up states:

> "ADR-030 D10 carries the reciprocal pointer, so whoever executes that survey sees this consumer without having to know ADR-036 exists."

ADR-030 v2 contains no reference to ADR-036 at all. The dependency is one-directional, and the document claims otherwise. This one matters beyond tidiness: ADR-036 D9's consult classifier is explicitly declared to **expire** when D10's typed-edge work lands, and D10 has since gained an illustrative `reliesOn` / `relyingParties` sketch that is exactly that survey's shape. The consumer ADR-036 relies on being findable is not findable.

**Also in scope for the pass**, since the same drift is likely:

- every component id ADR-036 references, against the component branch's corpus
- D9's consult-class label set against ADR-030 v2 D14, which moved the registry consult from the tool-side enforcement point to the agent-side one — that changes which arms are consult-class
- D4's aspect membership against the components the layer actually lands
- whether D9's "landing rule keyed on a rename-able label" survives the taxonomy renames in #452

**On the landing question**, which is settled and needs no ADR edit: the record lands on `main` as a tooling decision (ADR-002); the implementation (#464, #465) targets `develop` because it rewrites generated artifacts under `risk-map/`. Same split as ADR-030 v2 and its corpus layer.

### Done when

ADR-036 is in `docs/adr/` on `main` with `Status: Draft`, the `docs/adr/README.md` index row is present, and #464 and #465 cite a record a reviewer can read.

davidlabianca:
## [adr] File ADR-036 — decoupled component-graph emission

Closes: #472

Two files: the record and its index row.

### The problem

The component graph draws every edge flat. What makes it illegible is not the total edge count but the number of edges crossing a top-level category boundary — intra-cluster edges route almost free, cross-boundary ones force long spanning lines that tangle the rendering. No layout-engine setting reduces that count. Only changing which edges are drawn does.

### What it decides

**No edge is drawn between two top-level category clusters.** Three levers:

- **Aspects** — a pure sink many components write to is lifted out of the drawn graph and recorded in a header comment. Candidacy is mechanical: out-degree 0, plus a cross in-degree threshold.
- **Channels** — a cross-cluster edge becomes a labeled egress/ingress port pair, traced by matching labels rather than a line across the diagram.
- **Bands** — ports are pinned to their container's edge by subgraph nesting, so inbound ports line up in a column.

`emission.mode: flat | decoupled` is the rollback lever; the emergency-defaults path degrades to `flat`, so a corrupt config never yields a half-transformed diagram.

### What it looks like

[**Rendered decoupled graph**](https://raw.githubusercontent.com/davidlabianca/secure-ai-tooling/refs/heads/demo/decoupled-render/risk-map/svg/risk-map-graph.svg) — produced by flipping `emission.mode` on the implementation branch and regenerating through the repository's own generators, so it is reproducible rather than hand-drawn.

All three levers are visible at once: four clusters stacked with no lines between them, dashed port bands down each cluster's edge (`identity & authz ▸`, `▸ inference / serving ⇢ 2` — the `⇢ n` is the broadcast arm count), enforcement points in dashed outlines so a gate reads as a gate, and the lifted sink in amber marked `17 writes lifted`.

Measured on the same corpus:

| | flat | decoupled |
|---|---|---|
| **edges between top-level clusters** | **43** | **0** |
| total drawn edges | 90 | 84 |
| channel ports | 0 | 120 |

*(The render is on the implementation branch's corpus, which predates the component-layer rename — the sink node still reads "Secure Logging". The shape is what it illustrates.)*

### Scope: presentation only

No node and no `{to, from}` edge in `components.yaml` is added, removed, or retargeted. Every original edge survives as a drawn intra-cluster edge, is re-routed through a port pair, or is lifted into a header comment.

This is **not** the representation ADR that ADR-030 D10 defers — typed edge `kind`, directionality through a single enforcement point, control-intermediated flow views, overview-versus-detail composition. All four stay deferred; this re-routes the existing *untyped* edges without typing them.

### Two choices worth attention

**The aspect sink is named by role, not by component id.** `emission.aspects` is the normative registry; the record points there rather than restating it. That keeps a corpus rename a config edit instead of an erratum against this ADR.

**D9's consult-class set is two labels, with three as the promotion trigger.** A channel arm carrying a *consult* — a verdict the enforcement point consumes — lands on the component rather than the gate's inbound port, since drawing it into the gate reads as a second request stream. The classifier infers consult-ness from the concern label because the corpus cannot declare it, which holds only while every edge sharing a concern shares a class. Tests assert the bound: a third label fails rather than accumulates.

### What lands separately

This is the record only. The implementation and the config entries its registries govern land with the graph work, on their own branch — the record goes first so that change has something to be reviewed against.

### Reviewing

Status lands as `Draft` for a maintainer to flip.

Worth pushing on: whether aspect candidacy belongs at *sink-ness* (a sink receives copies and originates nothing, so lifting it hides no relationship — a source's fan-out *is* the structure the diagram exists to show), and whether the concern-label proxy is acceptable as a stopgap or should be a declared field now.


shrey-bagga:
**Review status:** ITERATE

### Summary

The ADR’s substantive design is coherent. One significant traceability claim remains false, D7’s normative port-ID grammar does not cover the multi-target IDs required by D6, and the branch currently conflicts with `main`.

### Required changes

#### [major] Add the reciprocal ADR-030 pointer that ADR-036 claims exists
- **Location:** `docs/adr/036-decoupled-component-graph-emission.md:212`
- **Problem:** ADR-036 says “ADR-030 D10 carries the reciprocal pointer.” The current ADR-030 v2 head (`92a15810457a22317bcf9a5ff2cd84092070325f`, PR #471) contains no reference to ADR-036 anywhere.
- **Why it matters:** ADR-036 describes this as a dependency rather than a suggestion: when typed edges land, D9’s label-based classifier must be replaced. Without the reciprocal pointer, someone executing ADR-030 D10 cannot discover that consumer as promised. Issue #472 also identifies this as a confirmed pre-filing defect.
- **Requested change:** Could we add the reciprocal dependency to ADR-030 v2 in PR #471 before finalizing this PR, or remove the assertion from ADR-036 and establish equivalent traceability explicitly?

#### [minor] Define the multi-target ingress port-ID grammar
- **Location:** `docs/adr/036-decoupled-component-graph-emission.md:80,92-94,110-115,128`
- **Problem:** D6’s normative example appends target suffixes—`_application` and `_reasoning_core`—when one root has multiple arms for the same concern. D7 defines only `p_in_<root>_<slug(concern)>`, which would give those arms the same ID.
- **Why it matters:** An implementation following D7 literally would collide Mermaid node IDs and fail to realize D6’s one-arm-per-distinct-target rule.
- **Requested change:** Could D7 define when the target suffix is required, while keeping the egress grammar unchanged?

#### [minor] Resolve the current main conflict
- **Location:** `docs/adr/README.md:54`
- **Problem:** GitHub reports the PR as conflicting because ADR-037 was added to the same index area after this PR’s base.
- **Why it matters:** The PR cannot merge in its current state, and resolving the conflict must preserve the intervening ADR status/index changes.
- **Requested change:** Could we rebase or merge current `main`, preserve ADR-034’s accepted status, and retain ADR-036 followed by ADR-037?

### Suggested documentation fixes

For D7:

```markdown
- **Content-derived port ids.** Egress ports use
  `p_out_<root>_<slug(concern)>`. Ingress ports use
  `p_in_<root>_<slug(concern)>` when that root has one arm for the
  concern; when the same `(root, concern)` has multiple distinct-target
  arms, append `_<slug(target)>`. Port ids are never positional
  (`p_in_0`…), so adding one component does not renumber unrelated ports.
```

For ADR-030 v2 D10:

```markdown
- **ADR-036 consumer.** [ADR-036](036-decoupled-component-graph-emission.md)
  D9 temporarily classifies consult arms from concern labels. When this
  representation lands a declared edge kind, that classifier must switch
  to the declared kind; the landing rule remains governed by ADR-036 D9.
```

### Checks run
- Reviewed PR #481 head: `4e3da4f55f2a478e16e0337201cd3609bf099b2a`
- Reviewed ADR-030 v2 PR #471 head: `92a15810457a22317bcf9a5ff2cd84092070325f`
- Reviewed component-layer PR #473 head: `88cb7d2887529b9be398ede7e93083f523156ce9`
- PR #473 corpus: 42 components, 96 directed edges, 45 cross-cluster edges
- `git diff --check`: passed
- `git merge-tree`: confirmed the `docs/adr/README.md` conflict
- GitHub checks: CLA passed; no other checks reported
- No executable tests run because this PR is documentation-only

davidlabianca:
Thanks @shrey-bagga  — both findings were real. Fixed in cb9220f. ptal

**[minor] D7's port-id grammar — fixed, with one correction to the suggested text.**

D7 now carries the conditional target suffix.

One change from your `diff`: `_<slug(target)>` would render `componentreasoningcore`, since `slug()` runs on the raw component id. D6's example shows `reasoning_core`. The suffix comes from `target_short_name(target)`, which strips the `component` prefix and splits camel case, so the bullet now reads `_<slug(target_short_name(target))>`. I kept your "across all channels" scoping — the implementation decides multi-arm per `(root, concern)` across every channel landing there, not within one channel, and that distinction is easy to lose on a re-read.

While in that bullet I also defined `<root>`, which it never did: it is the abbreviated category token (`infra`, `model`, `app`, `tools`), not D2's `root(n)`, which would give `componentsApplication` where D6's example uses `app`. Same defect class as the one you found, one line over.

On severity — you flagged that an implementation following D7 literally would collide node ids. Very true and the current implementation already computes the suffix conditionally and is correct but since that branch is local-only - you had no way to know.

**[major] The ADR-030 D10 reciprocal pointer — fixed, and the miss is ours.**

Confirmed: no version of ADR-030 mentions ADR-036. The Follow-up now says the pointer is *owed* rather than claiming it exists.

This one was on our list before filing — issue #472 records it as pre-filing defect #2, and the PR fixed defect #1 and shipped without this one. Thank you for catching it since we had already moved past it in error.

Adding the pointer to ADR-030 now needs its own PR against `develop` with a dated `## Amendment` section, since ADR-030 is Accepted — tracked as a follow-up.

**One thing your review surfaced that neither of us had named.**

D9's citations now state explicitly that ADR-030 v2 is currently on `develop`-only. Amending ADR-030 alone would not have fixed this, and could not be done from this PR.

**On your other two points.**

The `main` conflict was resolved by a rebase on 2026-08-20; the PR is mergeable, with ADR-037's index row present and ordered after ADR-036 and ADR-034 still Accepted.

Your corpus figures reproduce exactly — 42 components, 96 directed edges, 45 cross-cluster against #473's head. Neither `main` nor `develop` matches, which is expected: ADR-036:13 describes the corpus #473 lands and disclaims the counts as non-normative. Flagging it only because it is the kind of number a later reader might try to reproduce against `main` and conclude was wrong.

**Follow-ups from this review, none blocking here:**

- the reciprocal pointer into ADR-030 D10, as a dated amendment on `develop`
- a port-id **uniqueness** assertion in D7's self-check rubric — it currently checks per-band *label* uniqueness, which cannot catch an id collision, because colliding arms have distinct labels

shrey-bagga:
**Review status:** APPROVED

Thanks for the thorough response. I rechecked PR #481 at `cb9220f453a94042d0eee42ef775ee3b59a46e98` against its current `main` base:

- D7’s port-ID grammar now covers the multi-target arms required by D6, including the correct `target_short_name(target)` derivation.
- The rebase resolved the `main` conflict while preserving ADR-034 as Accepted and retaining ADR-037 after ADR-036.
- ADR-036 now accurately states that ADR-030 D10 does not yet contain the reciprocal pointer.

### Non-blocking follow-up

Current `main`’s ADR-030 D10 contains no ADR-036 reference, which is now consistent with ADR-036’s corrected Follow-up section. Could you please file or link the promised dated ADR-030 amendment so that this dependency remains discoverable after #472 closes?

The amendment can follow the repository’s appropriate branch and governance process separately; it does not block #481.

### Checks run

- PR head: `cb9220f453a94042d0eee42ef775ee3b59a46e98`
- Base checked: current `main` at `3a896b8a1b2c8f1949cf524a2f2bba349a9cc83a`
- GitHub mergeability: `MERGEABLE`
- CLA: passed
- `git diff --check`: passed
- No executable tests run because the PR is documentation-only

shrey-bagga (approved):

PR #483, opened 2026-08-05
Implements ADR-037 CI validation authority

### LINKED ISSUE #469: Enable --block on the CI validation jobs and cover the three ungated validators (ADR-037 D2, D4, D7)

davidlabianca:
Implements ADR-037
Closes #469

> **Reviewers — stacked on #470, which carries ADR-037 itself.** Until it merges, *Files changed*
> includes the ADR document. The implementation is the nine files surfaced by the following branch compare:
> https://github.com/davidlabianca/secure-ai-tooling/compare/chore/ci-block-flip...chore/ci-block-parity-impl

### What this does

ADR-037 D1: CI is the enforcing gate, and local may never be stricter than it. Four gaps closed.

**Strictness (D2, D4).** `--block` added to the two existing validation invocations, so a warn-level finding fails the build in CI as it already does in pre-commit.

**Coverage (D7, D8).** Nine validators that CI never ran gain invocations, one per validator so a failure names the check that failed. Three take `--block` and were simply never wired (D7). Six block unconditionally with no flag at all (D8). They run at their real paths rather than copied to the repository root: two derive `--schema-dir` from `Path(__file__)`, which a copy invalidates, after which field discovery returns silently and the job exits 0 having inspected nothing.

**Reachability.** `validation.yml` did not re-run when its own inputs changed — not on a validator edit, not on a `.pre-commit-config.yaml` edit, not on an edit to itself — and did not trigger on the content its validators *scan*. The vendor-neutrality jobs never ran on the agent and skill files they exist to check. `paths:` now covers each governed hook's own `files:` pattern plus each validator's import closure, derived rather than listed.

**Third-party hooks.** `check-jsonschema` and `check-metaschema` block locally and had no CI counterpart at all, which falsified D1's coverage rule rather than merely leaving a gap.

### Two production behaviour changes worth reading closely

Everything else is CI wiring. These two change what a hook gating every commit does.

**`_find_pairs()` now recurses.** It globbed `risk-map/schemas/*.schema.json` non-recursively, so it missed the archive pair — and that schema `$ref`s `riskmap.schema.json` three times. The fan-out hook, whose entire purpose is re-validating everything the master schema affects, was silently incomplete. Correct independently of CI parity.

**`_find_pairs()` now reads the index, not the filesystem.** `rglob` walks what is on disk, so an untracked scratch or backup directory under `risk-map/schemas/` blocked commits on files CI cannot see. It now filters through `git ls-files`, and fails loudly outside a repository rather than degrading to a filesystem walk — a silent fallback would reintroduce the defect under a condition nobody would notice. A *staged* pair is discovered, since that is what the commit in progress will contain.

### The part that is easy to get wrong

Three D7 validators take file arguments, and **two of the three exit 0 when handed no files**. A job that resolves an empty file list is a passing job that inspected nothing.

So the file list is not transcribed. `scripts/tools/hook_file_list.py` resolves it from each hook's own `files:`/`exclude:` regex over `git ls-files`, and exits non-zero rather than printing nothing when a pattern matches no file. Its output was verified byte-for-byte against pre-commit's own `Classifier.from_config().filenames_for_hook()`.

The same principle retired `validation.yml`'s hardcoded nine-target schema array in favour of the validator that derives its own pairs. Both checked the same nine; only one keeps doing so when a tenth schema is added.

### How this was tested

The suite asserts one invariant: **any edit that disables a check must turn it red.**

Forty-four mutations were run against it. The escapes clustered where nothing was looking:

- **Reachability.** Deleting the `pull_request:` trigger — or a `branches-ignore:` deny-list, or a `types:` filter naming no default event — left every assertion about steps and command lines intact while no gate ran on any pull request. `push:` is scoped to `main` and `develop`, so on a feature branch nothing ran at all.
- **Aggregation.** Deleting one `overall_success=false` line left `validation-summary` printing ❌ for a failed job and exiting 0 with "Overall Status: Success". The failure path is now *executed* — `${{ needs.* }}` expressions rendered the way GitHub does — rather than read.
- **Harness fidelity.** Step bodies ran in an empty directory, so a body shaped `if [ -d risk-map ]; then <success>; elif <validator>` took the branch it never takes in CI. The sandbox now carries the real tracked corpus.
- **Vacuity inside the non-vacuity tests.** The clean arm of one probe passed with a validator that discovered nothing.

Full suite: **3,512 passed, 15 skipped, 0 failed, 0 warnings.**

### On the size of this diff, and how to review it

This is a large review and I want to be upfront about why, because the shape is unusual: **7,912 of the 9,888 added lines are a single test file.**

`test_ci_block_parity.py` is one module rather than several because it asserts **one invariant over one surface** — the parity contract between `.pre-commit-config.yaml` and `.github/workflows/` — and every section shares the same derivations. The governed hook set, the gate-step set, the workflow trigger model and the validator import closure are each computed once at module scope and consumed by most sections. Splitting by section would either duplicate those derivations, which is how they drift apart, or create an import graph between test modules, which is worse. 

**The growth is a mutation record, not padding.** The suite claims that any edit disabling a check turns it red. Three adversarial passes ran 44 mutations against that claim; eleven survived green. Every survivor became a guard, and most guards carry a synthetic proof that they fail for the right reason. Roughly half the line count is that second layer.

Suggested reading order rather than top-to-bottom — the file is in numbered sections:

1. **Section 1, parser fidelity** (`:765`). Everything else quantifies over what these parsers return, so a bug here silently empties every downstream assertion. Read this one closely.
2. **Sections 2–4** (`:957`–`:1166`) are the ADR's D1/D3/D7b rules stated directly. Short, and they read like the decision record.
3. **Section 6, derived file lists** (`:1975`) and **section 7, step execution context** (`:3452`) are where the real defects lived. If you only read two sections, read these.
4. **Section 8, trigger coverage** (`:3874`) is the newest and the one I would most like a second opinion on — it reconciles a hook's `files:` regex against a workflow's `paths:` globs by sampling the tracked corpus rather than translating between the two languages.

The nine production lines that actually change behaviour are in `validate_all_schemas.py` and the two workflow files; `scripts/tools/hook_file_list.py` is the one genuinely new production surface at 232 lines. Those four files are the whole functional review. The rest is the argument that they cannot be silently disabled.

### What this composes with on `develop`

`--block` is a strictness flag, not a check. What it makes blocking is whatever the validator finds, and one such finding does not exist on either long-lived branch yet.

**#452** adds `check_category_style_coverage` to `validate_riskmap.py` — a component category with no Mermaid style entry. It targets `develop`; this targets `main`. Neither alone makes a missing category style fail CI:

| | `--block` in CI | category-style check | result |
|---|---|---|---|
| `main` today | no | no | not detected |
| `main` + this PR | yes | no | nothing to block on |
| `develop` + #452 | no | yes | detected, warn-only |
| after the next checkpoint merge | yes | yes | **blocking** |

Enforcement turns on at the `develop` → `main` checkpoint, not at either merge.

### Commits

Seven, each independently green, grouped by decision rather than by test-then-code.

| | |
|---|---|
| `a5ece1e` | brings CI validation to parity with the local gate |
| `3184cfe` | hardens the suite against edits that disable a check silently |
| `b1ca024` | covers the blocking validators that take no strictness flag |
| `36b6949` | replaces the transcribed schema list with the derived validator |
| `ab65930` | makes the gate reachable, attributable and non-vacuous |
| `eb458f7` | closes the mechanisms behind the escapes, not their instances |
| `4a74e67` | makes the last class-scoped fixture a staticmethod |

The last is unrelated to ADR-037: it is the repository's only class-scoped fixture defined as an instance method, a `PytestRemovedIn10Warning` today and a collection error under pytest 10, and the only warning the suite emitted. Refs #383 rather than closing it, since the fix belongs to that issue rather than to this decision.

### Scope

- Infrastructure → base `main` per ADR-002. No `risk-map/yaml/**` changes.
- ADR-037 D5's five copy-to-root steps are unchanged — measured on workflow YAML: zero `cp` line changes, six steps before and after.
- Stacked on **#470**, which should merge first.

### What reviewers should push on

- **`scripts/tools/hook_file_list.py` is the new production surface**, and the one place a silent bug would be invisible by construction. Its whole job is to prevent vacuity.
- **`validation.yml`'s `paths:` list stays explicit rather than a `scripts/**` glob**, so a Python-only change does not drag seven jobs including diagram rendering through CI. That is safe because the trigger-coverage test fails loudly on an uncovered dependency — but it is a deliberate trade.
- **The two `_find_pairs()` changes** alter a hook that gates every commit. The `$ref` justification for recursion and the fail-loud choice for the index filter are both worth a second opinion.
- **These jobs have never run.** The `paths:` fix means they will run on this pull request, which is their first real execution.

### DIFF
--- a/.github/workflows/validate_python.yml
+++ b/.github/workflows/validate_python.yml
@@ -6,12 +6,14 @@ on:
       - '.pre-commit-config.yaml'
       - 'requirements*.txt'
       - 'pyproject.toml'
+      - '.github/workflows/**'
   pull_request:
     paths:
       - '**.py'
       - '.pre-commit-config.yaml'
       - 'requirements*.txt'
       - 'pyproject.toml'
+      - '.github/workflows/**'
 
 permissions:
   contents: read
--- a/.github/workflows/validate_python.yml
+++ b/.github/workflows/validate_python.yml
@@ -7,6 +7,11 @@ on:
       - 'requirements*.txt'
       - 'pyproject.toml'
       - '.github/workflows/**'
+      # ADR-037 D1: test_ci_block_parity.py derives its governed-hook set from
+      # this ADR's instance table, so a PR editing only the table changes what
+      # the suite asserts without touching any workflow, script, or
+      # .pre-commit-config.yaml entry otherwise listed here.
+      - 'docs/adr/037-ci-validation-authority-and-block-parity.md'
   pull_request:
     paths:
       - '**.py'
@@ -14,6 +19,11 @@ on:
       - 'requirements*.txt'
       - 'pyproject.toml'
       - '.github/workflows/**'
+      # ADR-037 D1: test_ci_block_parity.py derives its governed-hook set from
+      # this ADR's instance table, so a PR editing only the table changes what
+      # the suite asserts without touching any workflow, script, or
+      # .pre-commit-config.yaml entry otherwise listed here.
+      - 'docs/adr/037-ci-validation-authority-and-block-parity.md'
 
 permissions:
   contents: read
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -15,6 +15,12 @@ on:
       - 'scripts/hooks/precommit/validate_identification_questions.py'
       - 'scripts/hooks/precommit/validate_yaml_prose_subset.py'
       - 'scripts/hooks/precommit/validate_prose_references.py'
+      - 'scripts/hooks/precommit/validate_mapping_purity.py'
+      - 'scripts/hooks/precommit/validate_mapping_drift.py'
+      - 'scripts/hooks/precommit/validate_versionid_purity.py'
+      - 'scripts/hooks/precommit/validate_all_schemas.py'
+      - 'scripts/hooks/precommit/validate_persona_site_build.py'
+      - 'scripts/hooks/precommit/validate_neutrality.py'
       - 'scripts/tools/hook_file_list.py'
   push:
     branches: [main, develop]
@@ -30,6 +36,12 @@ on:
       - 'scripts/hooks/precommit/validate_identification_questions.py'
       - 'scripts/hooks/precommit/validate_yaml_prose_subset.py'
       - 'scripts/hooks/precommit/validate_prose_references.py'
+      - 'scripts/hooks/precommit/validate_mapping_purity.py'
+      - 'scripts/hooks/precommit/validate_mapping_drift.py'
+      - 'scripts/hooks/precommit/validate_versionid_purity.py'
+      - 'scripts/hooks/precommit/validate_all_schemas.py'
+      - 'scripts/hooks/precommit/validate_persona_site_build.py'
+      - 'scripts/hooks/precommit/validate_neutrality.py'
       - 'scripts/tools/hook_file_list.py'
 
 permissions:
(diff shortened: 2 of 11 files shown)


shrey-bagga (approved):
**Review status:** APPROVED

I’m approving PR #483 at head `4a74e67df0a63799805adae8941efdc84d7bd4d9` based on the scope explicitly recorded in PR #470: **D9 is a decision, not an implementation, and is not an acceptance criterion for this stacked implementation.**

Under that agreed scope, PR #483 implements D1–D8 and covers the twelve current blocking-hook instances. 

### Checks run
- Focused tests: `414 passed, 9 skipped`
- Ruff and pre-commit configuration validation: passed
- All enforcing validator commands: passed
- GitHub checks: passing
- Live merge simulation against current `main`: clean

Issue #488, opened 2026-08-19
[bug] ComponentGraph's category container bypasses the subgraph-declaration guard, so a category and subcategory sharing an id emit a duplicate


davidlabianca:
## Summary

`risk-map/svg/risk-map-graph.svg` and its `.mermaid` source are generated by `ComponentGraph`. Its nested emission writes **two id namespaces into one flat space**, and only one of them passes through the duplicate guard. 

A duplicate `subgraph <id>` makes the whole file unrenderable — `mmdc` exits 1 with `TypeError: Cannot read properties of undefined (reading 'posX')`, and mermaid.live fails on it too.

**This is latent today, not live.** Category and subcategory enums are currently disjoint. The point of filing is that the corpus is growing into exactly the space where they meet.

## Mechanism

`base.py::_get_nested_subgraph_new` (`:506`):

```python
nested_subgraph.append(f'    subgraph {category} ["{category_name}"]')   # :513 — raw, unguarded
...
for subgroup_name, subgroup_components in category_subgroups.items():
    subgroup_lines = self._create_subgraph_section(subgroup_name, ...)   # guarded
```

- The **outer category container** is written with a raw f-string and **never registered**.
- The **inner subcategory declarations** go through `_create_subgraph_section`, which `f9bea51` made
  the single point that guarantees a declared id is not repeated.

The guard cannot see the category line - if a subcategory id ever equals a category id, the guard either renames the *subcategory* while the category keeps the original — silently splitting one logical box into two — or, where the category is emitted second, produces a genuine duplicate and the file stops rendering.

`base.py:473`'s own docstring already acknowledges that `_get_nested_subgraph_new` writes declarations directly. This issue is that acknowledgement turned into a defect report.

## Why now

The two namespaces are converging:

| | ids |
|---|---|
| subcategories, `main` today | `componentsData`, `componentsModelDeployment`, `componentsModelTraining`, `componentsModelCore`, `componentsOrchestration`, `componentsAgent`, `componentsApplicationCore` |
| subcategories after #473 | adds `componentsDeployment`, `componentsRegistries`, `componentsIdentity`, `componentsToolNetworkControls`, `componentsToolInvocationPath` |
| categories | `componentsInfrastructure`, `componentsModel`, `componentsApplication`, and `componentsTools` (added by #452) |

## Suggested fix

[...]

PR #490, opened 2026-08-19
fix(graphing): stop cluster-derived subgraphs colliding, vanishing, and renaming

davidlabianca:
Component clustering derives a subgroup id independently for each category, so two categories can land on the same candidate id. Nothing downstream expected that, and it failed four different ways — one loud, three silent.

### What was broken

**The loud one: the graph stopped rendering.** During MCP paper decomposition the same `subgraph` id was declared twice, which makes the whole Mermaid file unparseable — `mmdc` exits 1 with `TypeError: Cannot read properties of undefined (reading 'posX')`. A corpus growing controls from 37 to 67 produced it; renaming one occurrence by hand rendered the identical file.

**Silent component loss.** When a category's members were entirely absorbed into one cluster, the colliding entry overwrote the first, the emptied category was skipped, and its components disappeared from a graph that still rendered and still passed CI. The live corpus was **one component away** from that shape.

**A re-entrancy regression.** Collision state was scoped to the object rather than the render pass, and the risk graph reuses an already-built control graph — so every id in it was renamed on the second pass, producing unresolved edge targets and a box that lost its fill. No test caught it: the assertion was `"controlsData" in output`, and `controlsData2` satisfies it.

**A vacuous-truth coverage bug.** A category with no members made the full-coverage check trivially true — `set().issubset(x)` — so a control could appear to cover a category it had nothing to do with.

### What changed

- Declaration is now the single point that guarantees a cluster-derived id is written once. 
- Memberships merge rather than overwrite. 
- A category whose members are all absorbed is still visited, so its subgroup is emitted. 
- Collision state resets per render pass rather than per object. 
- A logical name rendered as two boxes is no longer offered as a coverage-optimisation target, since collapsing to one edge would point at one box and silently drop the other members.

### Generated artifacts are unchanged

All six regenerate byte-identical on the current corpus, where no collision occurs — verified through the real entry point (`validate_riskmap.py --force --to-graph … --to-controls-graph … --to-risk-graph …`), not by inspection. The fix only changes behaviour where a collision exists.

### How it was tested

Thirteen tests, 1,097 lines. They assert **guarantees rather than mechanisms**, because three separate patches to name generation left the defect untouched before the real cause was found downstream of naming — a monotonic counter produced four distinct names and the emitted graph still repeated the first.

Four of the assertions exist specifically because a rename-on-collision satisfies the naive check while leaving the output wrong: such a fix emits two `style` lines for the original id and none for the renamed subgraph, and leaves every edge pointing at the first. It renders, and it passes CI. Style-line count, edge-target resolution and component presence are what distinguish it from a real fix.

Each guard was verified by reproducing the mutation it catches, and the two review rounds this went through each rejected a version that was green.

### What this does *not* fix

**#488** — `ComponentGraph`'s category container writes its id raw and unguarded, so a category and
subcategory sharing an id still collide. Different mechanism, latent today (the enums are disjoint),
and tracked separately. The clustering machinery here is what that fix will need.

### Scope

Infrastructure → `main` per ADR-002. No corpus, schema or generated-artifact changes.

### DIFF
--- a/scripts/hooks/riskmap_validator/graphing/base.py
+++ b/scripts/hooks/riskmap_validator/graphing/base.py
@@ -69,6 +69,16 @@ class BaseGraph:
         self.controls: dict[str, ControlNode] = {}
         self.risks: dict[str, RiskNode] = {}
 
+        # Subgraph ids actually written by _create_subgraph_section, across
+        # every call made against this instance. Cluster naming (_find_node_
+        # clusters) derives a candidate id independently per call/category, so
+        # two different categories can legitimately derive the identical
+        # candidate (same common prefix, or same positional fallback index).
+        # _create_subgraph_section is the single place that turns a candidate
+        # into a declared "subgraph <id>" line, so it is the only place that
+        # can see -- and must guarantee -- that no id is declared twice.
+        self._declared_subgraph_ids: set[str] = set()
+
         self.component_by_category: dict[str, list[str]] = dict()
         self.component_by_subcategory: dict[str, dict[str, list[str]]] = dict()
         self.control_by_category: dict[str, list[str]] = dict()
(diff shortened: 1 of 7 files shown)


shrey-bagga (approved):
**Review status:** APPROVED

### Summary

This fix correctly prevents cluster-derived subgraph ID collisions and addresses the related component-loss, re-entrancy, and empty-category coverage paths. Collision handling is centralized, colliding memberships are preserved, ambiguous logical subgroups are excluded from single-edge optimization, and render state is reset at the correct boundaries. I found no actionable issues.

### Checks run

- Reviewed head: `f9bea51a99bd7218dc657b82c09bb5884317d978`
- Focused graphing tests: 145 passed
- Patch applied cleanly to current `main`; focused tests there: 145 passed
- Real graph regeneration: passed; all six generated `.md` and `.mermaid` files remained byte-identical
- Collision fixture rendered successfully with Mermaid CLI (`mmdc` exited 0)
- GitHub Python linting, testing, setup, validation summary, and CLA checks: passing

PR #497, opened 2026-08-25
Add candidate-value mode to audit-framework-mappings

### LINKED ISSUE #496: `audit-framework-mappings` can't verify a candidate mapping value before an entity exists
## `audit-framework-mappings` can't verify a candidate mapping value before an entity exists

Skill `audit-framework-mappings` audits mappings already resident in the corpus (a single entity, or the whole corpus) but has no first-class path for verifying a proposed mapping value for an entity that doesn't exist yet — the pre-PR authoring case `control-creator`/`risk-creator` need. Routing that case through the entity-scoped mode audits zero mappings and passes vacuously, even for a fabricated identifier.

Found while re-reviewing risk-creator/risk-critic, PR #435 against the shipped pattern for control-creator/control-critic., PR #434 — both instruct authors to verify candidate mapping values before finalizing a draft, and both inherited this gap.

### Scope

Add a third scope mode to the skill: audit a candidate value directly, keyed to a caller-stated entity type, with a four-step checklist (format/version, structural, selectivity, live-verify) so a candidate set can fail on over-selection or a technique/mitigation crossover, not just a malformed identifier.

davidlabianca:
Closes #496

The `audit-framework-mappings` skill audited mappings already resident in the corpus (single entity or whole-corpus) but had no first-class path for verifying a proposed mapping value on an entity that doesn't exist yet — the pre-PR authoring case. Routing that case through the entity-scoped mode audited zero mappings and passed vacuously, even for a fabricated identifier. This gap was inherited by `control-creator`/`control-critic` (already shipped) and `risk-creator`/`risk-critic` (in progress), both of which instruct authors to verify candidate mapping values before finalizing a draft.

## Changes

- Adds a third Scope mode: audit a candidate value directly, keyed to a caller-stated entity type, with an explicit halt-and-ask if that type is omitted.
- Adds a four-step candidate-mode checklist (format/version, structural, selectivity, live-verify), so a candidate set can fail on over-selection or a technique/mitigation crossover, not just a malformed identifier.
- Tightens the live-verify exemption to a closed-AND-unversioned test: STRIDE qualifies; ISO 22989, despite being a closed enum, does not (it's versioned and has a real external revision process); MITRE ATLAS/NIST AI RMF/OWASP/EU AI Act's open-ended identifier spaces never qualify.
- Adds 12 new eval cases (11-22) covering candidate mode across entity types, frameworks, and edge behavior (drip-fed candidate sets, entity type inferred from prose rather than a literal label, the soft-cap boundary).
- Updates the skill's frontmatter description and the `scripts/skills/README.md` roster entry so candidate mode is discoverable by a caller selecting skills by description alone.

## Known follow-ups (tracked separately, not in this PR)

- `risk-map/yaml/controls.yaml` maps `controlAgentPluginPermissions` to `AML.M0028@5.0.1`, which does not exist in the pinned MITRE ATLAS v5.0.1 edition — a real corpus content bug surfaced while building this fix's evals, unrelated to the skill itself.
- ADR-031's description of this skill ("verifies that a mapping's framework identifiers actually exist") predates the skill's whole-corpus live-verify carve-out; worth an architect-reviewed amendment to reconcile.
- ADR-027's drift-guard validators (`validate_mapping_drift.py`, `validate_mapping_purity.py`) are scoped to `risk-map/yaml/`/`risk-map/schemas/` and don't cover version-pinned identifiers embedded in `scripts/skills/**/evals/*.json`; a future framework-version bump won't flag stale eval fixtures.
- ADR-033 D7c's live-corpus eval floor isn't met for this skill's whole-corpus mode — none of the 22 eval cases exercises an actual whole-corpus sweep.

## Validation

- `python3 scripts/hooks/precommit/validate_neutrality.py` — clean.
- `evals.json` — valid JSON, 22 cases, ids 1-22 contiguous, uniform schema.

Co-authored-by: AI Assistant <ai-assistant@coalitionforsecureai.org>


shrey-bagga:
**Review status:** ITERATE

### Summary

This PR closes the gap from #496 by allowing mappings to be checked before a new risk or control has been added to the repository.

I found one important issue: some of the new evals approve a mapping without describing what the proposed risk is about. That means the skill can confirm that the mapping exists, but it cannot tell whether the mapping actually fits the risk.

There is also a smaller inconsistency about whether the skill should continue checking a mapping after an earlier check has already rejected it.

### Required changes

#### [major] Give the skill enough information to decide whether the mapping fits the risk

- **Location:** `scripts/skills/audit-framework-mappings/evals/evals.json:133-140, 160-169, 244-253` (new-file line numbers; eval IDs 13, 15, and 21)
- **Problem:** These evals provide mapping identifiers and expect the skill to approve them, but they do not describe the proposed risk. The skill can confirm that an identifier exists and is formatted correctly, but it has no information it can use to decide whether the mapping is directly relevant to the risk.
- **Why it matters:** A valid mapping is not automatically the right mapping for every risk. The skill’s checklist specifically requires it to check direct relevance, which means comparing what the mapping represents with what the proposed risk is about.
- **Requested change:** Could we add a short description of the proposed risk to these positive evals? If an eval intentionally leaves out that information, its expected response should say that the mapping passes the format, structure, and existence checks, but that the skill needs more context before deciding whether it is relevant.

#### [minor] Make the instructions and evals agree on whether checking can stop early

- **Location:** `scripts/skills/audit-framework-mappings/SKILL.md:61,85`; `scripts/skills/audit-framework-mappings/evals/evals.json:148,175` (new-file line numbers; eval IDs 14 and 16)
- **Problem:** The skill says that live verification always applies to candidate mappings, except for frameworks such as STRIDE. Evals 14 and 16 instead say that live verification is unnecessary once the mapping has already failed the structural check.
- **Why it matters:** This gives the skill two different instructions: always provide a live-verification result, or stop after an earlier check has already rejected the mapping.
- **Requested change:** Could we choose one behavior and make the instructions and evals match? If stopping early is intended, the expected response could explicitly say that live verification was not run because the mapping had already been rejected by the structural check.

### Questions

Could you help clarify what “current” means for a version-pinned mapping?

Should the skill confirm that the identifier existed in the version written in the mapping, confirm that it still exists in the latest framework version, or perform both checks?

ADR-027 allows mappings to remain pinned to an older recognized version, while ADR-031 also discusses checking whether an identifier was later retired or renamed. If both checks are intended, would you consider stating that explicitly and adding an eval that shows the difference?

### Checks run

- Reviewed PR head: `d9693b500343ac8ebff3d448e22a07632fe37ad1`
- `git diff --check`: passed
- Eval JSON parsing and structure: passed (23 evals)
- GitHub validation checks: passed
- Agent/Skill neutrality checks: passed
- CLA check: passed
- No dedicated eval runner was available, so the eval expectations were reviewed manually

shrey-bagga (approved):
We can take up these comments in a follow up issue

PR #499, opened 2026-08-25
chore: Remove the control and risk diagram generators

### LINKED ISSUE #477: [infra] Remove the control and risk diagram generators, keeping only --to-graph

davidlabianca:
## Remove the control and risk diagram generators

Closes: #477

Withdraws `ControlGraph` and `RiskGraph` and everything that existed only to serve them. The relationships those two diagrams carried remain available as tables — `controls-xref-components.md` and `controls-xref-risks.md` — which is where readers were already going, and which stay legible at a corpus size the diagrams had outgrown.

**59 files, +1,473 / −8,409.** All generated tables are byte-identical. The component graph loses exactly two lines — the `classDef hidden` and `classDef allControl` declarations, which were emitted into every graph and assigned to no node (see below); its node and edge content is unchanged.

### What goes

| | |
|---|---|
| Generators | `controls_graph.py`, `risks_graph.py` |
| CLI | `--to-controls-graph` / `-c`, `--to-risk-graph` / `-r`, and their `main()` blocks |
| Hook and CI | the `regenerate-graphs` trigger narrows to `components.yaml`; the CI graph loop collapses from three entries to one |
| Artifacts | 4 diagram files, 2 SVGs |
| Dead surface | `base.py` and `graph_utils.py` members reachable only from the removed generators, `MultiEdgeStyler`, `RiskNode`, `parse_risks_yaml`, the `mermaid-styles` control/risk config and its schema closure |
| Docs | 17 sites, rewritten to describe the component graph and to point at the surviving tables |

### Eleven commits — six removal, five closing the review finding

The first six are ordered by liveness and each is **independently green**, verified by running the suite at every boundary: 3753 / 3685 / 3697 / 3696 / 3695 / 3700.

| | | |
|---|---|---|
| `ae1bbb8` | remove the generators | 39 files |
| `90a9ea6` | sweep the surface that removal made dead | 13 files |
| `75d465e` | two review findings: an orphaned config knob, and a doc instructing a schema-rejected key | 7 files |
| `c7e5096` | a vacuous trigger guard, plus stale doc and ADR claims | 9 files |
| `51ecec3` | remove `cssClasses`, and two false ADR statements | 12 files |
| `2594568` | four small review leftovers: two tests, two doc corrections | 4 files |

The last five close this review's finding and follow the repository's red-green convention, so **four of them are deliberately failing** — they are the tests, landed before the fix that satisfies them. The tip is green at 3705.

| | | |
|---|---|---|
| `6e8bd9d` | RED — tests for the single-read contract | 17 failing |
| `f236be4` | RED — replace a `builtins.open` spy that a `Path.read_text` second read walks past | 17 failing |
| `b183d84` | RED — an audit-hook read counter and a non-vacuity assertion | 18 failing |
| `970834b` | RED — repair a stand-in that double-wraps once the contract ships | 18 failing |
| `1c1f610` | the implementation | green at 3705 |

The split is by *liveness*, not by file type: the first commit removes what is live, the second removes what the first made dead. That ordering is what lets each commit stand alone.

### Two findings worth a reviewer's attention

**The nesting check read the corpus twice, and the two reads could disagree.** Raised in review, and a requirement of #477 this branch had not met. `main()` re-opened `components.yaml` to reach its `categories:` block after `parse_components_yaml` had already read it, because that function returned components only and discarded the document. A corpus edited between the two reads produced manufactured nesting findings and exit 1 under `--block` on a corpus valid at every instant — reproduced by the reviewer, and now reproduced end to end against the fix, which exits 0.

`parse_components_yaml` now returns a frozen result carrying the raw document alongside the component mapping, both from one read; `main()` builds the category map from it. The validation pipeline reads the corpus exactly once. The graphing layer is out of scope and still re-reads it for category display titles, so a `--to-graph` run opens it twice — that read is fail-soft and cannot produce findings or change an exit code, but the claim is scoped rather than absolute.

**A pre-commit guard was vacuous, and this branch is what exposed it.** `test_precommit_hook_install.py` pinned the `regenerate-graphs` trigger with the string `(components|controls|risks)` but applied it via `re.search` — an alternation satisfied by any one branch matching anywhere. It stayed green when this branch narrowed the trigger to `components` alone, and would equally have accepted a widening or an unrelated path. It was the only test pinning that trigger. Both it and `regenerate-tables` now compare the whole value. `regenerate-svgs` is untouched: its pattern is backslash-escaped, so it never had this bug. That is not a claim its assertion is strong — it pins only the file extension, and the comment there now
says so rather than calling it correct.

**`sharedElements.cssClasses` was decorative before this branch.** Both its classes were emitted as `classDef` lines into every component graph and assigned to no node — the shipped artifact contains zero `:::` and zero `class` statements. Their assigner was `controls_graph.py`, so the emission had no effect even before #477; this branch removed the last pretext for it. Removing the config keys alone would have changed nothing, because `graph_utils` held hardcoded defaults identical to the YAML values and emitted them unconditionally — so the config, schema, fallbacks and emitters go together.

### ADR record

Three ADRs carried statements this work falsified:

- **ADR-019** and **ADR-020** listed `graphing/` as a downstream consumer of `risks.yaml` and `controls.yaml`. Neither is true after the sweep; corrected in place, as factual inventory rather than decision changes.
- **ADR-022** D6's substance changed, so it is amended under a dated `## Amendment` section with a new D9, a header pointer, and a status annotation in the ADR index — rather than edited in place.

### Verification

| | |
|---|---|
| Full suite at the tip | **3705 passed / 15 skipped / 0 failed** |
| The six removal commits, independently | 3753 / 3685 / 3697 / 3696 / 3695 / 3700 — all green |
| The single-read tests | killed three independently-built mutants: a `Path.read_text` second read, a read hoisted before the parse, and a renamed seam that made the guard fire on nothing |
| `--to-graph` `.md` + `.mermaid` | two `classDef` lines removed; node and edge content identical |
| `risk-map/tables/*` | byte-identical |
| Schema `$ref` closure | recomputed; zero orphans, zero dangling refs |
| `pre-commit run --all-files` | passes; tree unchanged |
| `ruff` | clean |

The schema trim was checked for permissiveness in both directions: the narrowed schema still rejects what it rejected before, and additionally rejects the two graph-type sections and the `subgroupFill` key it no longer defines.

### What is deliberately not here

- `parse_components_yaml` stays. It has live callers.
- The clustering machinery (`_find_component_clusters`, `_find_node_clusters`, `UnionFind`) stays. It has no production caller after this change, and is retained because the `ComponentGraph` category-collision fix needs it along with `f9bea51`'s guards. That retention is stated in the commit that would otherwise have removed it.
- Splitting the shared trigger-guard mechanism fix into its own change — that bug predates this branch and also covers `regenerate-tables`, which has nothing to do with graph removal.


shrey-bagga:
**Review status:** ITERATE

### Summary

Thanks for the thorough removal and follow-up cleanup. The focused test suite and all applicable GitHub checks pass. I found one requirement from #477 that still appears to be outstanding. I also noted a small PR-description nitpick.

### Required changes

#### [major] Parse `components.yaml` from one snapshot per run

- **Location:** `scripts/hooks/validate_riskmap.py:361-362` and `scripts/hooks/tests/test_validate_riskmap.py:914-975`
- **Problem:** The category/subcategory nesting check reopens `components.yaml` after `parse_components_yaml()` has already read it. The regression test still asserts exactly two opens and explains that #477 is expected to expose the parsed document and reduce this to one.
- **Why it matters:** The two reads can observe different file states during one validation run. I reproduced this by returning a different valid document from the second read; the validator opened the corpus twice, exited with status 1, and reported manufactured nesting/promotion failures.
- **Requested change:** Could we expose the raw parsed document alongside the component mapping from the original `yaml.safe_load()`, use that document for the nesting check, update affected callers, and change the regression test to assert one corpus open?

### Suggested code fix

One possible implementation shape is:

```python
@dataclass(frozen=True)
class ParsedComponents:
    document: dict[str, Any]
    components: dict[str, ComponentNode]


def parse_components_yaml(file_path: Path | None = None) -> ParsedComponents:
    if file_path is None:
        file_path = Path("risk-map/yaml/components.yaml")

    with open(file_path, "r", encoding="utf-8") as stream:
        document = yaml.safe_load(stream)

    # Extract the function's existing ComponentNode-building loop into this
    # pure helper so it consumes the already-loaded document.
    components = _build_components_from_document(document)
    return ParsedComponents(document=document, components=components)
```

The main validation path can then use the same result for both checks:

```python
parsed = validator.parse_corpus(components_file)

if not validator.validate_loaded(parsed.components, components_file):
    all_valid = False

category_to_subcategories = {
    category["id"]: {
        subcategory["id"]
        for subcategory in category.get("subcategory", [])
        if isinstance(subcategory.get("id"), str)
    }
    for category in parsed.document.get("categories", [])
    if isinstance(category.get("id"), str)
}
```

The exact return type is flexible; the important behavior is that component construction and the nesting check consume the same parsed document without reopening the file.

### Suggestions

Small nitpick: could we update the PR description before merge? It currently reports `+771`, while GitHub reports `+776`, and says `--to-graph` output is byte-identical. The generated Markdown and Mermaid outputs each lose the `hidden` and `allControl` `classDef` lines; commit `2866333` already describes that two-line change accurately.

### Checks run

- Reviewed head: `4df69c17544c7d97e26bd2ff292b71477a01c7b6`
- Focused exact-head test suite: 600 passed, 9 skipped
- Ruff on the changed Python surface: passed
- `git diff --check main...origin/pr-499`: passed
- GitHub checks: all applicable checks passed; deploy was skipped
- Runtime falsification check: confirmed two corpus opens and a manufactured blocking nesting/promotion result when the second snapshot differed

davidlabianca:
Thanks @shrey-bagga - all issues treated and the body of the PR captures the rework. ptal

shrey-bagga (approved):
Thank you so much for fixing the pending issues. Approved

Issue #401, opened 2026-07-06
Taxonomy: should agentic memory-consolidation (Reflection/Knowledge Consolidation) be a distinct lifecycle stage?


edonadei:
## Context

We were mapping risks and controls for an ADLC "Reflection / Knowledge Consolidation" phase and ran into a taxonomy gap. The 8 stages in [`risk-map/yaml/lifecycle-stage.yaml`](risk-map/yaml/lifecycle-stage.yaml) (`planning`, `data-preparation`, `model-training`, `development`, `evaluation`, `deployment`, `runtime`, `maintenance`) don't have a good fit for the point where an agent consolidates session or task output into durable memory.

The closest existing stage is `maintenance`. But `maintenance` is defined as "ongoing monitoring, updates, retraining, and system maintenance," which is system-level operations. The agent's per-session durable-memory write is a different thing, and mapping it onto `maintenance` loses that distinction.

The working notes and full risk/control breakdown for this phase live in the ADLC design doc: https://docs.google.com/document/d/16rb7h2y4FYu2tDsgVRA3HfUlXYEM0scVnQneCkB30SM/edit?tab=t.0

## Why this looks like a real gap

Some risks that seem specific to this phase have no home in the current map:

* Self-Generated False Memory Consolidation. An agent persists an incorrect-but-unpoisoned conclusion from its own reasoning into durable memory. `riskDataPoisoning` presumes an adversarial external data source, and `riskRetrievalVectorStorePoisoning` presumes an external corpus. Neither covers the agent calcifying its own reasoning.
* Durable-Knowledge Provenance Loss. Knowledge is saved with no source attribution. `riskAgentDelegationChainOpacity` covers provenance of actions, not provenance of consolidated knowledge.
* Memory-Write Approval-Gate Bypass. A human-approval gate on high-impact durable-memory writes is bypassed, misconfigured, or degraded. `riskRogueActions` is the nearest neighbor, but it addresses unintended actions rather than the failure of this control.

Separately, three risks that plausibly belong to this phase currently carry no `lifecycleStage` tag at all: `riskPromptResponseCachePoisoning`, `riskRetrievalVectorStorePoisoning`, and `riskCovertChannelsInModelOutputs`. The "poisoned or covert content promoted into durable memory" angle has no stage to attach to.

## Questions for maintainers

[...]

Issue #170, opened 2026-03-18
[Update Persona]: ISO 22989 uses "AI Customer" instead of "AI Consumer"


bact:
### Persona Permalink*

https://github.com/cosai-oasis/secure-ai-tooling/blob/acc260a2b65d5506a5f76b674abdc241acd17275/risk-map/yaml/personas.yaml#L139-L153

https://github.com/cosai-oasis/secure-ai-tooling/blob/acc260a2b65d5506a5f76b674abdc241acd17275/risk-map/yaml/personas.yaml#L170-L184

### Change Type*

Other (describe below)

### Proposed Changes*

Current state:

- `personaApplicationDeveloper` maps to "AI Consumer (application builder)"
- `personaEndUser` maps to "AI Consumer (end user)"

Proposed change:

- `personaApplicationDeveloper` maps to "AI Customer (application builder)"
- `personaEndUser` maps to "AI Customer (end user)"

Rationale:

ISO/IEC 22989:2022 Clause 5.19 (AI stakeholder roles) uses the term "AI customer" (Clause 5.19.4).

### Framework Impact Assessment

Impact analysis:

- This change does not change interpretation of existing controls.

### Supporting Evidence

- ISO/IEC 22989:2022 Information technology — Artificial intelligence — Artificial intelligence concepts and terminology https://www.iso.org/standard/74296.html

### Submission Checklist

- [x] I have provided a valid GitHub permalink to the persona

PR #500, opened 2026-08-26
feat(lifecycle): add decommissioning as lifecycle stage 9

edonadei:
**Review this:** 5 files, small. About 5 minutes.

**State:** ready for review as of 2026-09-16. All checks green at `5a2cd49`; this head adds only docs and one description paragraph. The blocker named in the earlier body (RFC [ws4#170](https://github.com/cosai-oasis/ws4-secure-design-agentic-systems/issues/170) disposition) is cleared: the reviewers on that thread converged on 9/3 to 9/14 and @imran-siddique cleared it here on 9/9. The RFC body edit and the WS4 chair acceptance are still in progress on ws4#170; nothing in that edit changes this diff.

**What this makes work:** `decommissioning` becomes a taggable lifecycle stage, so risks about retiring an agent stop landing on `maintenance`.

## What changes

1. `risk-map/schemas/lifecycle-stage.schema.json`: add `decommissioning` to the id enum, move the `order` maximum from 8 to 9.
2. `risk-map/yaml/lifecycle-stage.yaml`: add the stage entry at order 9, update the header count.
3. `risk-map/docs/guide-metadata.md`: the two stage counts, and the "Valid Values" list (nine entries now).
4. `risk-map/docs/design/metadata-mappings.md`: the "8-Stage AI Lifecycle Model" heading and list, flagged in review. Now nine.

Additive only. No existing stage is renamed, reordered, or removed.

## Why

The corpus lifecycle ends at `maintenance`. Risks whose exposure is specific to *retiring* an agent (authenticator revocation, delegation-tree teardown, memory and trace disposition) have no stage to tag. Today they land on `maintenance`, which covers ongoing operation of a live system, or they go untagged.

Implements RFC [ws4#170](https://github.com/cosai-oasis/ws4-secure-design-agentic-systems/issues/170).

## Decisions already baked in

The stage is **terminal**. Decommissioning ends the agent's authority to act; pause, suspend, quarantine and kill switch stay in `runtime` or `maintenance`; any later return is a new admission. This is the definition ws4#170's reviewers settled on (@NicolaiNielsenTR, @imran-siddique, @raymondsheh, @billbrietstout), replacing the original proposal's soft/hard split rather than renaming it.

The wording follows those reviews. Authenticator revocation and identity tombstoning rather than identity deletion, because deleting an identifier breaks replay detection. Disposal against a verified disposition record rather than "deletion", since a conforming outcome may be retained under hold or crypto-shredded. And, from the 9/3 review, the terminal state and its effective time are published where a relying party can verify them, so a record signed after retirement is distinguishable from one signed before.

## Two things reviewers usually ask

`order` carried `"maximum": 8`, so a ninth stage fails validation on the numeric bound as well as on the id enum. Both move to 9.

`docs/adr/022-supporting-schemas.md` says 8 in four places. It is a historical ADR recording a decision at a point in time, so this PR leaves it alone, per the 9/9 review. `scripts/hooks/tests/test_lifecycle_stage_order_uniqueness.py` has a stale fixture comment ("8 stages"); fixtures are inline and pass, and `scripts/**` bases on `main` per ADR-002, so it stays out of this PR.

## Landing order

Taxonomy prerequisite (ADR-034 D1). Sequence after this merges:

1. Companion PR: append `decommissioning` to five existing risks (`riskShadowAndUnknownAgents`, `riskZombieShadowMCPServers`, `riskSensitiveDataDisclosure`, `riskExcessiveDataHandlingDuringInference`, `riskRetrievalVectorStorePoisoning`).
2. Follow-on catalog PR: new decommissioning risk IDs, from ws4#170's follow-on table.

Independent of the reflection stage ([#401](https://github.com/cosai-oasis/secure-ai-tooling/issues/401)); numbering reconciles if both advance. The observability component PR (#501) referenced in the earlier body is closed; its successor is a control, [ws4#195](https://github.com/cosai-oasis/ws4-secure-design-agentic-systems/issues/195), and does not interact with this stage.

## Validation

1. `check-jsonschema --schemafile lifecycle-stage.schema.json ../yaml/lifecycle-stage.yaml`: ok
2. `python3 scripts/hooks/validate_riskmap.py --mode lifecycle --force`: passed
3. `pytest scripts/hooks/tests/test_lifecycle_stage_order_uniqueness.py`: 20 passed

---

*PR body and the 2026-09-16 revision drafted with Claude Fable 5.1; reviewed by @edonadei.*

### DIFF
--- a/risk-map/docs/guide-metadata.md
+++ b/risk-map/docs/guide-metadata.md
@@ -347,7 +347,7 @@ The schemas enforce these validation rules for metadata fields:
 
 3. **Enum Constraints**: All specific values must match their respective schema enums:
    - Framework IDs in `mappings` must exist in `frameworks.yaml`
-   - Lifecycle stages must be one of the 8 stages defined in `lifecycle-stage.schema.json`
+   - Lifecycle stages must be one of the 9 stages defined in `lifecycle-stage.schema.json`
    - Impact types must be one of the 10 types defined in `impact-type.schema.json`
    - Actor access levels must be one of the 9 levels defined in `actor-access.schema.json`
 
@@ -357,7 +357,7 @@ The schemas enforce these validation rules for metadata fields:
 
 1. **Be selective**: Only include metadata fields that add meaningful context
 2. **Use multiple values**: Most metadata fields accept arrays - include all relevant values
-3. **Consider the full lifecycle**: Think about where risks/controls apply across all 8 stages
+3. **Consider the full lifecycle**: Think about where risks/controls apply across all 9 stages
 4. **Map to multiple frameworks**: Cross-referencing helps users from different security backgrounds
 5. **Choose primary impacts**: Focus on the most significant impact types rather than listing everything
 
--- a/risk-map/schemas/lifecycle-stage.schema.json
+++ b/risk-map/schemas/lifecycle-stage.schema.json
@@ -28,7 +28,8 @@
             "evaluation",
             "deployment",
             "runtime",
-            "maintenance"
+            "maintenance",
+            "decommissioning"
           ]
         },
         "title": {
@@ -41,9 +42,9 @@
         },
         "order": {
           "type": "integer",
-          "description": "Sequential order in the lifecycle (1-8)",
+          "description": "Sequential order in the lifecycle (1-9)",
           "minimum": 1,
-          "maximum": 8
+          "maximum": 9
         }
       },
       "required": ["id", "title", "description"],
--- a/risk-map/yaml/lifecycle-stage.yaml
+++ b/risk-map/yaml/lifecycle-stage.yaml
@@ -5,10 +5,11 @@
 title: Lifecycle Stages
 description:
   - >
-    Defines the phases of AI system development, deployment, and operation. This eight-stage
+    Defines the phases of AI system development, deployment, and operation. This nine-stage
     lifecycle covers the complete journey from initial planning through data preparation,
-    model training, development, evaluation, deployment, runtime operation, and ongoing
-    maintenance. Each stage presents unique security considerations and risk profiles
+    model training, development, evaluation, deployment, runtime operation, ongoing
+    maintenance, and eventual decommissioning. Each stage presents unique security
+    considerations and risk profiles
 lifecycleStages:
   - id: planning
     title: Planning
@@ -49,3 +50,11 @@ lifecycleStages:
     title: Maintenance
     description: Ongoing monitoring, updates, retraining, and system maintenance
     order: 8
+
+  - id: decommissioning
+    title: Decommissioning
+    description: >-
+      Terminal retirement of the AI system: revoking authenticators and tombstoning
+      identities, revoking delegated authority across descendants, and disposing of
+      models, data, memory, and decision traces against a verified disposition record
+    order: 9


edonadei:
this is clearly not ready to be filed yet, I'll update it when the RFC will get more feedback

imran-siddique (commented):
Endorsing this, and clearing what the PR body names as its blocker.

**On the block.** The body says this is held until ws4#170 is disposed. From my side that is done: on 3 September I took @edonadei's own sentence on #170 as the definition, that decommissioning ends the agent's authority to act, and that a return re-enters through Admission as a new entity. Dropping soft versus hard mode entirely is better than what I argued for on 20 August, which was removing soft mode and keeping the surrounding structure. So the definitional question this PR was waiting on is settled as far as my review goes, and I would take it out of draft.

**What I checked in the diff.** The change is additive: no existing stage id is renamed, reordered or removed, `order` runs 1 to 9 with the schema maximum moved to match, and the enum gains one value. The stage text carries both things #170's reviews asked for, authenticator revocation with identity tombstoning rather than identity deletion, which keeps replay detection working, and disposal stated against a verified disposition record rather than as deletion. Terminal is the right call and it is what both reviews reached independently. All 28 checks green at the current head, including `license/cla`.

**One thing to fix before or with merge.** `risk-map/docs/design/metadata-mappings.md` line 211 has a heading "8-Stage AI Lifecycle Model" over an enumerated list of all eight stages, sourced from `lifecycle-stage.yaml`. Nothing in this diff touches it, so on merge it becomes a second, shorter definition of the lifecycle sitting next to the real one. Two lines: the heading, and a ninth entry.

`docs/adr/022-supporting-schemas.md` also says 8 in four places (lines 92, 99, 272, 273, including the `order` 1-8 range), but that is a historical ADR recording a decision at a point in time and I would leave it alone rather than retro-edit it. Editors' call.

Nothing else blocking from me.

edonadei:
Out of draft, per the 9/9 review.

`51c6d33` addresses the one fix you asked for and one more of the same kind:

- `risk-map/docs/design/metadata-mappings.md`: "8-Stage" heading and list, now nine.
- `risk-map/docs/guide-metadata.md`: the "Valid Values" list under Lifecycle Stage was still eight entries. The two counts further down were already updated in the first commit; the list was not.
- `risk-map/yaml/lifecycle-stage.yaml`: the stage description now carries the definition ws4#170 settled on after your 9/3 comment: ends the agent's authority to act, terminal state and effective time published for relying parties, any return is a new admission. No id or order change.

ADR-022 left alone, as you suggested. Requesting @davidlabianca for the Risk Map side.

davidlabianca:
@edonadei thanks for the proposal, and for working through the ws4#170 thread.

This PR adds a new value, `decommissioning`, to `lifecycleStage`. Before we accept a new metadata value, we need a proposed use of it in the existing corpus. It can be partial, but it has to exist.

**Why:** a value that no entry uses is content debt. The schema accepts it and the docs describe it, but nothing is tagged with it. Nobody has committed to applying it, so the work falls to whoever comes next, and there's no record of which entries it was meant for.

**To move this forward:**
- Open an issue that references this PR and proposes the first corpus update for `decommissioning`.
- A table is enough: the risk or control ID, whether it **adds** `decommissioning` or **moves** from `maintenance` to it, and a one-line rationale. The PR body names the risk classes it has in mind: authenticator revocation, delegation-tree teardown, and disposal of memory and decision traces. A good place to start is the 7 risks and 3 controls currently tagged `maintenance`.
- Components don't carry `lifecycleStage`, so the table only needs risks and controls.

I'm moving this PR back to draft and deferring it until that issue is in place. Once it is, we'll pick this up again and the two can land together.

edonadei:
Thank you for the context, in that updated rfc there's that block that mention potentially two risks that could apply

_[...] **riskShadowAndUnknownAgents** and **riskZombieShadowMCPServers** are clear candidates. [...]_

Though, I did let that PR rot a little bit too much, the RFC 170 has been closed in favour of an updated one https://github.com/cosai-oasis/ws4-secure-design-agentic-systems/issues/205, whenever I will file the issue you mention, I will also clean up that PR.

cc @cngo20 @getglad co-chairs for the ADLC SIG

Issue #195, opened 2026-04-07
[Risk]: Insufficient Logging and Auditability of Delegation Chains


billbrietstout:
### Risk ID*

LDC

### Risk Title*

Insufficient Logging and Auditability of Delegation Chains

### Short Description*

Agent actions and multi-hop delegations lack immutable, correlated logs, making it impossible to reconstruct who (human or agent) authorized or performed a given action.

### Long Description*

Agentic systems rely on On-Behalf-Of (OBO) delegation and multi-hop tool calls via MCP. Insufficient Logging and Auditability of Delegation Chains occurs when logs fail to capture the full actor chain, correlation IDs, or delegation metadata in an immutable and queryable way. Even when individual authentication succeeds, the provenance of actions becomes opaque.

This vulnerability manifests through missing or incomplete logging at the IAM, orchestration, or MCP layers — for example, omitting the original caller, intermediate agents, or token exchange steps. Attack vectors include deliberate suppression by compromised agents, logging gaps in dynamic tool chaining, or reliance on volatile logs that do not preserve delegation context.

The potential impact includes inability to investigate incidents, attribute actions for compliance or forensics, detect anomalous delegation patterns, or perform effective revocation. Consequences involve accountability failures, regulatory violations, prolonged breach dwell time, and loss of trust in autonomous systems.

This risk is a core concern in the CoSAI WS4 Agentic Identity and Access Control document (loss of actor clarity and broken delegation chains) and is directly addressed in MCP-T12 (Insufficient Logging, Monitoring, and Auditability) of the Model Context Protocol Security document. It becomes especially critical in regulated or high-stakes environments using agentic AI.

**Proposed Risk Category: risksApplicationAndAgentic**

### Risk Category*

risksDeploymentAndInfrastructure

### Applicable Personas*

- [ ] personaModelProvider
- [ ] personaDataProvider
- [ ] personaPlatformProvider
- [ ] personaModelServing
- [x] personaAgenticProvider
- [x] personaApplicationDeveloper
- [ ] personaGovernance
- [ ] personaEndUser
- [ ] personaModelCreator
- [ ] personaModelConsumer

### Applicable Controls*

controlModelAndDataIntegrityManagement
controlImmutableDelegationLogging (proposed)

### Relevant Questions (Optional)

Do logs capture the full delegation chain with immutable correlation IDs for every action?
Can you replay and attribute any agent-tool interaction after the fact?
[...]

PR #501, opened 2026-08-26
feat(components): add Observability with semantic and behavioral sub-components

edonadei:
**Answer this one question and the PR is unblocked:** is Observability a top-level category, or a subcategory under Application? Everything else is done.

**State:** draft. All 5 corpus checks green. Prerequisite RFC filed as [ws4#175](https://github.com/cosai-oasis/ws4-secure-design-agentic-systems/issues/175). Reviewable at the 2026-08-26 SIG meeting ([ws4#174](https://github.com/cosai-oasis/ws4-secure-design-agentic-systems/discussions/174)). Review time about 10 minutes.

**What this makes work:** Observability exists as a component with 2 sub-components, wired to `componentReasoningCore` with reciprocal edges, so the 8 observability risks in PR 4 have somewhere correct to attach.

## The open decision

`components.schema.json` closes `category` to four values. The SIG proposal calls Observability a new *top-level category*, which would widen that enum and also touch the nesting `allOf`, the issue-template dropdown, and the graph grouping.

This branch implements the narrower reading: `componentsObservability` as a subcategory under `componentsApplication`, which keeps it beside the agent components it instruments.

Both are defensible. A top-level category argues that observability spans the agent, orchestration, and model layers rather than sitting inside the application. Tracked as open question 1 in [ws4#175](https://github.com/cosai-oasis/ws4-secure-design-agentic-systems/issues/175). This PR stays a draft until it is settled, because the answer changes the diff.

## What changes

1. `components.schema.json`: add 2 component ids, add `componentsObservability` to the subcategory enum, add it to the `componentsApplication` branch of the nesting `allOf`.
2. `components.yaml`: add the subcategory definition and the 2 component entries with their edges.
3. `components.yaml`: add reciprocal edges to `componentReasoningCore`.

## Why observability needs its own component

The corpus has no component representing the instrumentation that captures, protects, and analyses agent decisions. The nearest existing component, `componentReasoningCore`, is the *subject* being observed rather than the *observer*. The core produces decisions; observability records them and evaluates them against expected behaviour.

Around fifteen observability risks across the ADLC phases are currently force-filed under codes that do not fit:

| Observability risk | Force-filed under | Why it is wrong |
| --- | --- | --- |
| Sensitive data in decision traces | `riskSensitiveDataDisclosure` | Covers the inference path, not the logging path |
| Non-tamper-evident trace storage | `riskModelDeploymentTampering` | Covers artifact substitution, not log integrity |
| Trace suppression or deletion | `riskModelDeploymentTampering` | Log *availability* is distinct from deployment integrity |
| Incomplete behavioural baseline | `riskRogueActions` | Covers the outcome of undetected drift, not the missing instrument |

Each misfiling buries the observability concern and overloads a code that was never scoped to carry it.

Three properties make this agent-specific rather than classical logging. Decision traces are a novel aggregation point, concentrating inputs, outputs, and intermediate reasoning into a store with weaker access controls than the sources those values came from. Non-determinism makes the trace load-bearing for telling agent error apart from agent compromise. And suppression is unrecoverable, because the trace is the only record of non-deterministic reasoning.

## Why two sub-components

They carry distinct risk profiles and neither is sufficient alone. Semantic Observability *produces* the record: traces, reasoning logs, delegation-chain events, provenance. Behavioral Observability *consumes* it, covering drift and anomaly detection plus the intervention path of circuit breakers, kill switches, and escalation gates.

## Two things reviewers usually ask

Edges are reciprocal, so this touches the Reasoning Core. `validate_riskmap.py` enforces that if A declares `to: B`, then B declares `from: A`. `componentReasoningCore` therefore gains the matching pair: the core feeds Semantic Observability, and Behavioral Observability feeds back into the core along the intervention path. Per ADR-034 D3, a new component lands with its edges, never edgeless.

Does #473 already cover this? #473 adds 17 components. Its ADR-030 v2 dependency (#471) merged 2026-08-20, so it is waiting on its own review rather than on an external blocker. If #473 already contains these two sub-components as scoped here, this PR should be withdrawn in its favour. If it is a partial match, the two need reconciling before either lands.

## Landing order

Layer 1 per ADR-034: a referenced-only leaf landing with its edges, before any risk or control references it. The eight follow-on observability risks are Layer 5 and land in a separate PR after this one. Independent of the decommissioning stage PR (#500).

## Generated artifacts

Regenerated by the repo's own hooks: the risk-map and controls diagrams (`.md` and `.mermaid`), the three SVGs, the component tables, and the `new_component` issue-template dropdown.

## Validation

1. `check-jsonschema --schemafile risk-map/schemas/components.schema.json risk-map/yaml/components.yaml`: ok
2. `python3 scripts/hooks/validate_riskmap.py --force`: component edges consistent, plus lifecycle order, controls↔components mirror, category/subcategory nesting, and category style all passed


edonadei:
The RFC from that PR still needs to be settled before going forward https://github.com/cosai-oasis/ws4-secure-design-agentic-systems/issues/175

edonadei:
Closing without merge. The RFC this implements, cosai-oasis/ws4-secure-design-agentic-systems#175, is withdrawn: `componentAuditRecordRepository` already covers semantic observability, and behavioral observability is being refiled as a control in cosai-oasis/ws4-secure-design-agentic-systems#195 (*Agent Behavioral Monitoring*) rather than as a component.

Keeping the branch: the schema and validator plumbing here is reusable for the control PR once #195 clears its three-day window.

PR #506, opened 2026-08-27
Fix mapping-selection’s tool-generate marker and NIST/EU AI Act mapping gaps

davidlabianca:
The `mapping-selection` skill's `[tool-generate]` inline marker convention breaks the pinned-pattern regexes it's meant to flag (they're fully anchored and reject any suffix), and a dual-review pass surfaced four more correctness/coherence gaps while fixing it.

## Changes

- Retires the `[tool-generate]` marker in favor of omit-and-flag, matching `audit-framework-mappings`'s established convention, and routes mapping-value verification through that skill's single-entity or candidate mode depending on whether the entity has a corpus row yet.
- Replaces the NIST AI RMF illustration `GOVERN-6.2` with `MEASURE-2.7` — `GOVERN-6.x` is scoped to third-party/supply-chain issues only, and two real corpus controls (`controlAgentPluginPermissions`, `controlAgentCredentialIsolation`) already misapply it to internal permission/credential mechanics; the eval now also asserts against that specific misread.
- Scopes the "deliberately deferring a value" declaration to candidate mode only — single-entity mode has no drip-feed mechanism to route it to.
- Clarifies that a single-entity "complete set" submission is scoped per framework, and adds a default for a caller who cannot yet say whether a further value is genuinely coming.
- Adds EU AI Act to the control-side framework enumeration, which previously implied the framework never applies to anything this skill covers.

## Known follow-ups (tracked separately, not in this PR)

- Whether single-entity mode's "complete set" requirement (bundling existing corpus mappings with a newly proposed value) is actually needed by `audit-framework-mappings` — its own output format suggests it looks up an entity's current mappings itself — is an open design question, not resolved here.
- `controlAgentPluginPermissions` and `controlAgentCredentialIsolation` both carry `GOVERN-6.2@1.0` on internal permission/credential mechanics, which this skill's own reference file says that subcategory should never cover — a pre-existing corpus content bug, unrelated to this skill.

## Validation

- `python3 scripts/hooks/precommit/validate_neutrality.py` — clean on all three touched files.
- `evals.json` — valid JSON, 14 cases, ids 1-14 contiguous, uniform schema.

Co-authored-by: AI Assistant <ai-assistant@coalitionforsecureai.org>


shrey-bagga (approved):
Review status: APPROVED
This update cleanly resolves the mapping-selection gaps. Removing the [tool-generate] marker prevents invalid mapping values, while the revised workflow correctly separates drafting candidates from candidate-mode or single-entity verification. The NIST AI RMF and EU AI Act guidance is clearer, and the added evals cover the new behavior.
I reviewed head 27b89a635b3ada6a1057a2639a7276c17ac6fb37. 

All 18 GitHub checks and the focused local validations pass. I found no blocking issues.

PR #523, opened 2026-09-01
merge: Periodic sync of `main` into `develop` (01-sep-2026)

davidlabianca:
Brings `main` (`b42eef7`) into `develop`. **27 main-only commits, 67 files, 10 deletions** — principally #499, which removed the control and risk graph generators per #477.

## Why this one needed more than a merge

Git flagged **9 conflicts**, all mechanical. The work was in the **5 carriers it did not flag** — each invisible to conflict detection for a different reason:

| carrier | why git was blind |
|---|---|
| `test_generation_tools_category_smoke.py` | develop-only file; `main` cannot conflict with a file it lacks |
| `test_components_yaml_tools_category.py` | `parse_components_yaml`'s **return type** changed, not its name |
| `conftest.py` | auto-merged clean; fixture kept keys the narrowed schema now rejects |
| `graph_utils.py` *(production)* | schema key nothing reads, so no test observed it |
| `docs/adr/036-*.md` | a `path:line` citation — correct when written, wrong after the merge moved the file |

## Conflict resolutions

- **6 modify/delete** on the artifacts #499 deleted → take the delete. Nothing survives that the tables lack.
- **`validate_riskmap.py`** — kept develop's `MermaidConfigLoader` imports, took main's removal of `ControlGraph`/`RiskGraph`/`parse_risks_yaml`. Re-read end to end; the previous backmerge shipped a `NameError` through the same file and the same shape of conflict.
- **`mermaid-styles.yaml`** — kept develop's `componentsExternalTools` category, took main's `subgroupFill` removal. Schema-validates.
- **`risk-map-graph.svg`** — **regenerated, not selected.** Taking either side reverts one branch's change; the committed SVG had carried CSS its own source can no longer produce.

## Also repaired

`graph-customization.md` instructed contributors to author a `sharedElements` block the merged schema rejects — `main` correctly trimmed the example to its three categories while `develop` added a required fourth. Neither side was wrong alone. Two unresolvable SHA citations in ADR-022 dropped in favour of the `#477` issue link already in the same sentence.

## Verification

`pytest scripts/hooks/` **0 failed** · `ruff check` clean · `pre-commit run --all-files` all hooks pass, tree undirtied · component graph/`.mermaid`/SVG regenerate byte-identical · every `path:line` citation in the 25 changed documents resolves.


davidlabianca:
Periodic backmerge with small true-ups...

